PR #281, opened 2021-08-16
Added the 143 validation rule.

### LINKED ISSUE #32: Error code 143 - The reason for new episode code is not a valid code.
#### Error code
143
#### Description to display
The reason for new episode code is not a valid code.
If completed reason for new episode must be coded S, P, L, T, U or B.
#### DfE suggested coding
If provided `<RNE>` must be a valid value  For allowed values please see associated code set.

jfan2006:
Added the reason for new episode code checking for 143 validation rule. 

Closes #32


tests/test_validators.py
```diff
@@ -1,6 +1,19 @@
...
+        'RNE': ['S', 'p', 'SP', 'a', '', pd.NA],
+    })
+
+    fake_dfs = {'Episodes': fake_data}
+
+    error_defn, error_func = validate_143()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'Episodes': [1, 2, 3, 4, 5]}
```
mwaddoups: There's actually a subtlety to this rule that needs to be covered in the test and code (according to DfE). 

The description specifies "If provided <RNE> must be a valid value.".  Note the "if provided" - this means, that if it's not provided (i.e. it's pd.NA) , it shouldn't be a validation error. 

As such, I think this test case should just return `[1, 2, 3, 4]` as index 5 is valid.

validator903/validators.py
```diff
@@ -1,6 +1,30 @@
...
+    )
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        
+        episodes = dfs['Episodes']
+        code_list = ['S', 'P', 'L', 'T', 'B']
+
+        mask = episodes['RNE'].isin(code_list)
```
mwaddoups: Similarly to my comment on the test, we need to also allow for cases where it is not provided. In this case you will want to combine this mask with another mask using the OR (`|`) operator. The relevant functions in pandas you'll need to check to see if it is missing are `.notna()` and `.isna()`.

See if you can get the test to pass by changing this.

mwaddoups (changes requested):

tests/test_validators.py
jfan2006: This is interesting! Totally understood your explanation about the DfE spec but I'm thinking in reality there shouldn't and won't any episode that doesn't have this reason code. (If it's before the end of June I'd try this on the DfE's SSDA903 system) 

I wondered I should contact DfE to double check/ clarify this validation rule... but for this project I'll make the change.

mwaddoups: Interesting - my thought is that you might see it missing for unclosed episodes at the end of the year? But that's off the top of my head, I can't remember for sure.

jfan2006: I think you were thinking about rule 148 - that's the reason episode ceased - the reason episode ceased could be blank (together with date episode ceased) if they were open/ unclosed episodes at the year end.

> Interesting - my thought is that you might see it missing for unclosed episodes at the end of the year? But that's off the top of my head, I can't remember for sure.

mwaddoups: Yep - you're right!

mwaddoups (approved):
All looks good now - nice work! Approving.

PR #285, opened 2021-08-19
Added validation for Error code 141

### LINKED ISSUE #30: Error code 141 - Date episode began is not a valid date.
#### Error code
141
#### Description to display
Date episode began is not a valid date.
If completed date episode began must be a valid date and should be entered in the format DD/MM/YYYY.
#### DfE suggested coding
If provided `<DECOM>` must be in a valid format and contain a valid date  Format should be DD/MM/YYYY.

WMossDCC:
Validation added to ensure DECOM date on Episodes is valid DD/MM/YYYY.

Closes #30


tests/test_validators.py
```diff
@@ -12,4 +12,17 @@ def test_validate_101():
 
     result = error_func(fake_dfs)
 
-    assert result == {'Header': [2, 3]}
\ No newline at end of file
+    assert result == {'Header': [2, 3]}
+
+def test_validate_102():
```
mwaddoups: This test is misnamed (so may actually not be being ran!)

tests/test_validators.py
```diff
@@ -12,4 +12,17 @@ def test_validate_101():
...
+        'DECOM': ['01/01/2021', '19/02/2010', '38/04/2019', '01/01/19', pd.NA],
+    })
+
+    fake_dfs = {'Episodes': fake_data}
+
+    error_defn, error_func = validate_141()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'Episodes': [2, 3, 4]}
```
mwaddoups: The text of this validation actually hides a subtlety - it mentions "If provided, the date must be a valid value". This suggest to me that if it's missing it should actually pass (so index 4 should be valid)

I think in practice the DECOM will usually always be present, but its worth coding to handle the edge case.

mwaddoups (changes requested):
Looks good overall but a couple small things to sort out

validator903/validators.py
```diff
@@ -24,6 +24,29 @@ def _validate(dfs):
...
+    )
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        else:
+            episodes = dfs['Episodes']
+            mask = pd.to_datetime(episodes['DECOM'], format='%d/%m/%Y', errors='coerce').notna()
+
+            NA = episodes['DECOM'].isna()
```
mwaddoups: Change to `na_locations`

mwaddoups (approved):
All looks good - approved!

PR #292, opened 2021-08-20
Added errors 388, 213, 214, 222, 366   (hopefully)   to validators an…

PhilDeakin:
Added errors 388, 213, 214, 222, 366 to validators and config.  I feel I may have messed up the indentation in validators hopefully I'm wrong. Thanks.

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -194,3 +194,101 @@ def validate_392c():
             return {'Episodes': episodes.index[error_mask].tolist()}
 
     return error, _validate
+
+def validate_388():
+    error = ErrorDefinition(
+        code='388',
+        description='Reason episode ceased is coded new episode begins, but there is no continuation episode.',
+        affected_fields=['REC'],
+    )
+
+    def _validate(dfs):
+        if 'Episode' not in dfs:
+            return {}
+        else:
+            df = dfs['Episode']
+            error_rows = (df['CHILD'] != df['CHILD']) 
+            for x in df['CHILD'].unique():
+              #Form a loop through each unique child id
+              this_child = df[df['CHILD'] == x].sort_values('DECOM')
+              #Focus on each child in turn (thischild)
+              if len(this_child['CHILD']) >= 2:
+              #Check this child has at least 2 rows (otherwise it's fine)
+                for inchild in range(len(this_child['CHILD']) - 1):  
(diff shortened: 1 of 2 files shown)


validator903/validators.py
```diff
@@ -194,3 +194,101 @@ def _validate(dfs):
...
+        else:
+          df = dfs['Episode']
+          mask = (df['PLACE'].isin(['H5','P1','P2','P3','R1','R2','R5','T0','T1','T2','T3','T4','Z1'])) & (df['URN'].notna()) & (df['URN'] != 'XXXXXX')
+          return {'Episode': df.index[mask].tolist()}
+
+    return error, _validate
+
+
+
+def validate_366 ():
```
mwaddoups, line 279: Try to avoid these spaces here - it technically works I think but better to keep it consistent (i.e. `def validate_366()`)

validator903/validators.py
```diff
@@ -194,3 +194,101 @@ def _validate(dfs):
...
+        code='366',
+        description='A child cannot change placement during the course of an individual short-term respite break.',
+        affected_fields=['RNE'],
+    )
+
+    def _validate(dfs):
+        if 'Episode' not in dfs:
+            return {}
+        else:
+          df = dfs['Episode']
```
mwaddoups, line 290: As you mentioned, in a few spots the indentation is inconsistent - for this project it would be better to just stick to 4 spaces between levels.

I appreciate repl.it makes this a bit harder than it should be but it should pick up on the 4 spaces once you start writing code that way.

mwaddoups (changes requested):
Hey Phil - thanks for uploading all of these and good to see lots of work! However there are a few larger things to check on in the work here.

One of the key parts is the tests - but I'm not seeing any tests uploaded here. Not sure if it's accidental but I think I've spotted a couple things which I'd expect tests to have caught.

I would suggest you just open a single pull request for one of the simple rules (e.g. 213) and just make sure you have done every step in the Contribution Guide - it could be worth rewatching the final video as well. I think once you get that down for a couple PRs then we can jump into larger PRs for 4 or 5 rules.

I'd suggest leaving this PR open as a draft and creating some new ones for a couple of the easier rules together with tests.

I've noted a few minor things in the comments too.

I can see you've grasped loads of the concepts really quickly which is brilliant - just need to follow the process a bit more to get it into the right form and working!

mwaddoups:
Closing as all of these are now completed elsewhere!

Issue #275, opened 2021-07-29
Error code 633 - Local authority code where previous permanence option was arranged is not a valid value.


mwaddoups:
#### Error code
633
#### Description to display
Local authority code where previous permanence option was arranged is not a valid value.
#### DfE Long-form description of error
The local authority code should be a valid local authority code or one the following codes:  NIR = Northern Ireland, NUK = outside of the UK, SCO = Scotland, WAL = Wales.
#### DfE suggested coding
Where provided `<LA_PERM>` must be a valid value

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

mwaddoups:
This one is complicated and requires more work - see conversation on #295

tab1tha:
Hi @SLornieCYC @PhilDeakin 
This issue would require a significant amount of time to implement so we are evaluating to see whether or not to prioritise it at this moment. Could you please help by answering a few questions?

- Do you use the LA_PERM field?
- What values do you put into the field (old 3 digit code? new long code?)
- Is it common to have this error (633).

Context: LA_PERM will have to be compared with the list of local authority codes and we are thinking through how that list should be gotten, stored and possibly updated.

PhilDeakin:
@tab1tha 
Do you use the LA_PERM field? Yes, occasionally, but generally to enter our own LA code. Extremely rare that we put a different LA's code (maybe one in the past 5 years)

What values do you put into the field (old 3 digit code? new long code?) 3 digit (the guidance seems to suggest it should be the 3 digit code again this year too)

Is it common to have this error (633). No, I don't recall us having any errors of that type.

tab1tha:
I see. Thank you @PhilDeakin .

tab1tha:
One more question: is it possible to derive the 3 digit codes from the long codes? @PhilDeakin 
We have the long codes already (but the short codes are required, as you said).

PhilDeakin:
@tab1tha I'm not sure sorry

Issue #273, opened 2021-07-29
Error code 631 - Previous permanence option not a valid value.


mwaddoups:
#### Error code
631
#### Description to display
Previous permanence option not a valid value.
#### DfE Long-form description of error
Previous permanence option not a valid value.  Valid values are:  P1 = Adoption, P2 = Special guardianship order (SGO), P3 = Residence order (RO) or child arrangements order (CAO), P4 = unknown, Z1 = Child has not previously had a permanence option.
#### DfE suggested coding
If provided `<PREV_PERM>` must be a valid value  For allowed values please see associated code set.

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

Issue #95, opened 2021-07-29
Error code 1009 - Reason for placement change is not a valid code.


mwaddoups:
#### Error code
1009
#### Description to display
Reason for placement change is not a valid code.
#### DfE Long-form description of error
CARPL, CLOSE, ALLEG, STAND, APPRR, CREQB, CREQO, CHILD, LAREQ, PLACE, CUSTOD or OTHER are the only valid entries for reason for placement change.
#### DfE suggested coding
If provided `<REASON_PLACE_CHANGE>` must be a valid value  For allowed values please see associated code set.

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

Issue #92, opened 2021-07-29
Error code 1006 - Missing type invalid.


mwaddoups:
#### Error code
1006
#### Description to display
Missing type invalid.
#### DfE Long-form description of error
Missing type should be either M - missing from care or A - away from placement without authorisation.
#### DfE suggested coding
If provided `<MISSING>` must be a valid value  For allowed values please see associated code set.

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

PR #295, opened 2021-08-21
Added some rules related to if completed must be from a list - 114, 1…

### LINKED ISSUE #19: Error code 114 - Data entry to record the status of former carer(s) of an adopted child is invalid.
#### Error code
114
#### Description to display
Data entry to record the status of former carer(s) of an adopted child is invalid.
If the child has been adopted during the year, the code must be either:  0 = Child not adopted by former foster carer(s), 1 = Child adopted by former foster carer(s).
#### DfE suggested coding
If provided `<FOSTER_CARE>` must be a valid value  For allowed values please see associated code set.

jfan2006:
Added some validation rules related to 'if completed must be from a list' - 114, 120, 131, 132, 175, 176, 177, 196, 631, 633, 1006, and 1009.

Closes #19 , #25 , #26 , #27 , #52 , #53 , #54 , #83 , #273 , #275 , #92 , and #95 .


validator903/validators.py
```diff
@@ -1,6 +1,311 @@
...
+        return {'Missing': validation_error_locations.tolist()}
+
+    return error, _validate
+
+def validate_633():
+    error = ErrorDefinition(
+        code='633',
+        description='Local authority code where previous permanence option was arranged is not a valid value.',
+        affected_fields=['LA_PERM'],
+    )
```
mwaddoups: I would omit this error for now.

If you read the description carefully, what it says is that the LA code *must be a valid code or* one of the following codes.

What the rule actually wants is for the value to be checked against a list of local authority codes in England and Wales. Having looked at previous permanence files before, LA_PERM is often left blank - so not 100% sure how to deal with this one.

Would be good to get insight from yourself/other analysts on what your previous permanance files look like and if the value is filled in.

For now, I would just remove this rule from this PR - I've removed the tag 'simple rule'.

validator903/validators.py
```diff
@@ -1,6 +1,311 @@
...
+
+        return {'PrevPerm': validation_error_locations.tolist()}
+
+    return error, _validate
+
+def validate_196():
+    error = ErrorDefinition(
+        code='196',
+        description='Strengths and Difficulties (SDQ) reason is not a valid code.',
+        affected_fields=['SDQ_REASON'],
```
mwaddoups, line 90: This isn't a change - more just a note on this one.

The full text specifies to only check this if a child has been looked after continuously for 12 months or more. However, my understanding is that will be true for all children who have OC2 data - so I think I agree we don't need to check that. Does that match your understanding as well?

validator903/validators.py
```diff
@@ -1,6 +1,311 @@
...
+    )
+
+    def _validate(dfs):
+        if 'AD1' not in dfs:
+            return {}
+        
+        adoptions = dfs['AD1']
+        code_list = [1, 2]
+
+        mask = adoptions['NB_ADOPTR'].isin(code_list) | adoptions['NB_ADOPTR'].isna()
```
mwaddoups: For this one, it's not clear that NB_ADOPTR will always come in as an integer column rather than a string column. If it's a string, it may create a weird bug where for example the string `'1'` is not in the list `[1, 2]`.

To be on the safe side here, it would be worth
- Adding a test to make sure it validates correctly for all of `[1, 2, '1', '2']`.
- Altering the code - I would suggest doing something like `adoptions['NB_ADOPTR'].astype(str).isin(['1', '2'])`, but there are other ways to solve it too.

validator903/validators.py
```diff
@@ -1,6 +1,311 @@
...
+          'F2', 
+          'P2', 
+          'F3', 
+          'P3', 
+          'G4',
+          'G5', 
+          'G6', 
+          0
+          ]
+        mask = care_leavers['ACTIV'].isin(code_list) | care_leavers['ACTIV'].isna()
```
mwaddoups: Again, for this one I would use the string `'0'` rather than the number `0`, and make sure you convert the type of ACTIV correctly.

I would also add a special test case for the `'0'`  value.

validator903/validators.py
```diff
@@ -1,6 +1,311 @@
...
+    )
+
+    def _validate(dfs):
+        if 'AD1' not in dfs:
+            return {}
+        
+        adoptions = dfs['AD1']
+        code_list = [0, 1]
+
+        mask = adoptions['FOSTER_CARE'].isin(code_list) | adoptions['FOSTER_CARE'].isna()
```
mwaddoups: And again, the same comment on this one as with 175
- Make sure you handle strings / integers
- Add test cases for both as well.

tests/test_validators.py
```diff
@@ -1,6 +1,162 @@
...
+        'FOSTER_CARE': [0, 1, '0', '1', 2, 'former foster carer', '', pd.NA],
+    })
+
+    fake_dfs = {'AD1': fake_data}
+
+    error_defn, error_func = validate_114()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'AD1': [2, 3, 4, 5, 6]}
```
mwaddoups: 2 and 3 should pass here (as mentioned in my comment on the main function). Since all this data is CSV we can't distinguish between strings and numbers.

tests/test_validators.py
```diff
@@ -1,6 +1,162 @@
...
+        'NB_ADOPTR': [0, 1, 2, '1', '', pd.NA],
+    })
+
+    fake_dfs = {'AD1': fake_data}
+
+    error_defn, error_func = validate_175()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'AD1': [0, 3, 4]}
```
mwaddoups: 3 should pass here (as mentioned in other comments)

mwaddoups (changes requested):
Wow - lot's of new stuff here and a big leap forward!

One minor issue around comparing strings/numbers comes up in a few of these rules that needs minor changes.

And one of the rules should be removed from this PR as it actually has a much more difficult solution.

Nice work so far!

validator903/validators.py
jfan2006: You're right - I have just gone back a few years and see all the files they are all indeed blank.

validator903/validators.py
jfan2006: @mwaddoups that's right - OC2 only contains those continuously in care for 12 months. 
Refer to the page 96 '2.9.2 Children covered by this return' of [the DfE guide](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974676/Children_looked-after_by_local_authorities_in_England_2020_to_2021.pdf).

jfan2006:
Have made changes for all requested. 

Removed rule 633; 
amended validator for rules 175, 132, 114 and their tests to make sure both string and integer work for allowed value.

validator903/validators.py
jfan2006: @PhilDeakin Phil, what's your understanding around this? As we only use CSV files, I wondered LA_PERM are all blank in XML  files too?

mwaddoups (approved):
Nice one Fan - great to see this many rules coming in! Thanks for making the minor changes.

validator903/validators.py
PhilDeakin: @jfan2006 The vast majority are null in our CSVs, but there are some that have Manchester's LA code 352.

PR #296, opened 2021-08-21
Added validation rule 178.

### LINKED ISSUE #55: Error code 178 - Placement provider code is not a valid code.
#### Error code
178
#### Description to display
Placement provider code is not a valid code.
If completed placement provider of each episode must contain a code from the range PR0 to PR5.  The placement provider code should be left blank if the record is only recording the young person's activity and accommodation after leaving care.  The placement provider code should be left blank for children in any of the following placements: T0, T1, T2, T3, T4 or Z1.
#### DfE suggested coding
If provided `<PLACE_PROVIDER>` must be a valid value and should not be provided if `<PL>` = 'T0'; 'T1'; 'T2'; 'T3' or 'Z1'.  For allowed values please see associated code set.

jfan2006:
Added validation rule 178.

Closes #55


validator903/validators.py
```diff
@@ -1,6 +1,30 @@
...
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        
+        episodes = dfs['Episodes']
+        code_list_1 = ['PR0', 'PR1', 'PR2', 'PR3', 'PR4', 'PR5']
+        code_list_2 = ['T0', 'T1', 'T2', 'T3', 'T4', 'Z1']
+
+        mask = episodes['PLACE_PROVIDER'].isin(code_list_1) & ~episodes['PLACE'].isin(code_list_2) | episodes['PLACE_PROVIDER'].isna() | episodes['PLACE_PROVIDER'].isna() & episodes['PLACE'].isin(code_list_2)
```
mwaddoups: So this code works 👍 but I'd suggest breaking this into a few more lines to make the code a bit more readable and maybe changing some variable names to help that too - just to make it easy for anyone who has to glance at or tweak this in the future.
- `code_list_1` -> something like `place_provider_codes`
- `code_list_2` -> something like `place_codes_with_no_provider`

Then for the mask, maybe break into a line for each case, e.g. `place_provider_needed_and_correct`, `place_provider_not_provided`, `place_provider_not_needed` and combine it all together with ORs at the end.

mwaddoups (changes requested):
Code all works well - but I just had some suggested minor changes around readability.

jfan2006:
@mwaddoups I have made the changes as requested.

mwaddoups (approved):
This all looks good now Fan - nice work!

PR #303, opened 2021-08-24
Added Error 388 using new logic, no looping

### LINKED ISSUE #103: Error code 388 - Reason episode ceased is coded new episode begins, but there is no continuation episode.
#### Error code
388
#### Description to display
Reason episode ceased is coded new episode begins, but there is no continuation episode.
If reason episode ceased is coded X1 there must be a subsequent episode starting on the same day.  If an episode ends but the child continues to be looked after, a new episode should start on the same day. The reason episode ceased code of the episode which ends must be X1.  If a child ceases to be looked after reason episode ceased code X1 must not be used.
#### DfE suggested coding
If `<REC>` = 'X1'  then next episode `<DECOM>` must be = `<DEC>`

PhilDeakin:
As above...


PhilDeakin:
Clears issue #103

PhilDeakin:
This is a stupid question, but as soon as a new section of code is committed to main, does it start working on the website immediately or do you have to do something further to make it work on the website?

mwaddoups:
I currently have to manually build and copy the code so it doesn't immediately start working - I've been trying to do it relatively often, there may be a bit of a delay at the moment as have been working on a few new features.

We hopefully should automate it soon!

tests/test_validators.py
```diff
@@ -185,3 +185,19 @@ def test_validate_168():
     result = error_func(fake_dfs)
 
     assert result == {'Header': [1,3,4,6,7,9,10,11,12,16]}
+
+def test_validate_388():
+    fake_data = pd.DataFrame({
+        'CHILD': ['111','111','111','111','111','111','111','111','111','222','333','333','333','333','444'],
+        'DECOM': ['01/01/2021','02/01/2021','03/01/2021','04/01/2021','05/01/2021',pd.NA,'07/01/2021','08/01/2021','09/01/2021','10/01/2021','11/01/2021','12/01/2021','13/01/2021','14/01/2021','15/01/2021'],
+        'DEC': ['02/01/2021','03/01/2021','04/01/2021','05/01/2021','06/01/2021','07/01/2021','08/01/2021','09/01/2021',pd.NA,pd.NA,'12/01/2021','13/01/2021','14/01/2021','15/01/2021',pd.NA],
+        'REC': ['X1','X1','X1','E11','X1','X1','X1','X1',pd.NA,pd.NA,'X1','X1','E12','X1',pd.NA],
```
mwaddoups: I think it would be good to add some test cases to cover the "REC is X1, but the next episode starts on a different date then the end of the last episode" case - I don't think this is covered in the list here?

I would also cover the "X1 must not be used if care has ceased" - so for example, a child with a single episode that has ended cannot have the code X1.

Quite a few cases to cover in testing with this rule!

mwaddoups: I'd also add a test case for garbage data (i.e. DECOM or DEC are in the wrong format or missing) - we don't want the rule to crash if the file is still in the early stages of validation and doesn't have a valid DECOM

validator903/validators.py
```diff
@@ -383,3 +383,25 @@ def _validate(dfs):
...
+            return {}
+        else:
+            df = dfs['Episode']
+            df['DECOM'] = pd.to_datetime(df['DECOM'], format='%d/%m/%Y', errors='coerce')
+            df['DEC'] = pd.to_datetime(df['DEC'], format='%d/%m/%Y', errors='coerce')
+            df = df.sort_values(['CHILD','DECOM'])
+            df['LAG_DECOM'] = df['DECOM'].shift(-1)
+            df2 = df.loc[~df.index.isin(df.groupby('CHILD')['DECOM'].idxmax()),:]
+            df2 = df2[(df2['DEC'] == df2['LAG_DECOM']) & (df2['REC'] != 'X1')]
+            return {'Episode': df2.index.tolist()}
```
mwaddoups: I think your code here only covers the case where episodes follow each other but the code is not X1.

However, I think an episode where the code is X1 but the next episode doesn't have the right DECOM would not work. I have suggested adding more of these test cases in the tests.

validator903/validators.py
```diff
@@ -383,3 +383,25 @@ def _validate(dfs):
...
+        if 'Episode' not in dfs:
+            return {}
+        else:
+            df = dfs['Episode']
+            df['DECOM'] = pd.to_datetime(df['DECOM'], format='%d/%m/%Y', errors='coerce')
+            df['DEC'] = pd.to_datetime(df['DEC'], format='%d/%m/%Y', errors='coerce')
+            df = df.sort_values(['CHILD','DECOM'])
+            df['LAG_DECOM'] = df['DECOM'].shift(-1)
+            df2 = df.loc[~df.index.isin(df.groupby('CHILD')['DECOM'].idxmax()),:]
+            df2 = df2[(df2['DEC'] == df2['LAG_DECOM']) & (df2['REC'] != 'X1')]
```
mwaddoups: Good use of the new functions - I think it would be good to split the code out over a few more lines and use more descriptive variable names, as this would improve readability (and maintainability). Some suggestions:
- In general, it's good to separate different logical blocks of code with blank lines. Since there's quite a lot going on here, you can consider splitting the datetime/sort from the indexing - up to you to decide the details.
- Maybe consider 'DECOM_NEXT_EPISODE' rather than 'LAG_DECOM'
- Pull the `groupby` onto it's own line and call it something like `last_episode_indexes`
- Use a more descriptive name than `df2`.
- Consider pulling out the indexing expressions as well to make it easy for e.g. a reader who might forget what X1 means specifically.

Comments can also help with readability - but I think just choosing good variable names and layout will get you there without needing to add comments too.

tests/test_validators.py
mwaddoups: Also - it may help readability for the test if you build the test dataframe in a different way (using record syntax) as the test data is quite large and difficult to track across the different lists. You could use
```
fake_data = pd.DataFrame([
    {'CHILD': '111', 'DECOM': '01/01/2021', 'DEC': '02/01/2021', 'REC': 'X1'},
    {'CHILD': '111', 'DECOM': '02/01/2021', 'DEC': '03/01/2021', 'REC': 'X1'},
    #... etc
])
```

mwaddoups (changes requested):
Good work on reading up on the more complex stuff and using the ideas to good effect here.

However I think as this is quite a complex rule there are still some cases that are missing - I've tried to list some here but I may well have forgotten stuff. It's important that the tests cover all the corner cases and are as exhaustive as possible.

Once you update the tests with all the right cases, the rest of the code changes should be straightforward I think.

validator903/validators.py
```diff
@@ -383,3 +383,62 @@ def _validate(dfs):
...
+        affected_fields=['REC'],
+    )
+
+    def _validate(dfs):
+        if 'Episode' not in dfs:
+            return {}
+        else:
+            df = dfs['Episode']
+            df['DECOM'] = pd.to_datetime(df['DECOM'], format='%d/%m/%Y', errors='coerce').fillna(pd.to_datetime(df['DECOM'], format='%d/%m/%y', errors='coerce'))
+            df['DEC'] = pd.to_datetime(df['DEC'], format='%d/%m/%Y', errors='coerce').fillna(pd.to_datetime(df['DEC'], format='%d/%m/%y', errors='coerce'))
```
mwaddoups: I may be missing something but I don't see a need for the `fillna` calls on these two lines? I see you're handling a different date format, but I think it would likely make more sense to just treat anything which doesn't match the expected format as a row we won't validate.

validator903/validators.py
```diff
@@ -383,3 +383,62 @@ def _validate(dfs):
...
+
+            df['DECOM'] = df['DECOM'].fillna('01/01/1901')
```
mwaddoups: There might be a better way to handle this by just ignoring episodes where `df['DECOM'].isna()` - I'm always a bit wary of fill values as they might create hard to spot issues down the road (like say if they introduced a new code for "end of episode less than 30 days" or something).

However, thinking about this is hard and I think if this is working there is no big need to change this - up to you.

tests/test_validators.py
```diff
@@ -185,3 +185,36 @@ def test_validate_168():
...
+    { 'CHILD' : '6666', 'DECOM' : '12/06/2020', 'DEC' : '13/06/2020', 'REC': 'X1' },  #11
+    { 'CHILD' : '6666', 'DECOM' : '13/06/2020', 'DEC' : '14/06/2020', 'REC': 'X1' },  #12
+    { 'CHILD' : '6666', 'DECOM' : '14/06/2020', 'DEC' : '15/06/2020', 'REC': 'X1' },  #13
+    { 'CHILD' : '6666', 'DECOM' : '15/06/2020', 'DEC' : '16/06/2020', 'REC': 'X1' },  #14  Fails Case 3
+    { 'CHILD' : '77777', 'DECOM' : '16/06/2020', 'DEC' : '17/06/2020', 'REC': 'X1' }, #15
+    { 'CHILD' : '77777', 'DECOM' : '17/06/2020', 'DEC' : '18/06/2020', 'REC': 'X1' }, #16
+    { 'CHILD' : '77777', 'DECOM' : '18/06/2020', 'DEC' : pd.NA, 'REC': pd.NA },       #17
+    { 'CHILD' : '999', 'DECOM' : '31/06/2020', 'DEC' : pd.NA, 'REC': pd.NA },  #18   Nonsense date, but should pass
+    { 'CHILD' : '123', 'DECOM' : pd.NA, 'DEC' : pd.NA, 'REC': pd.NA },  #19   Nonsense dates, but should pass
+    { 'CHILD' : pd.NA, 'DECOM' : pd.NA, 'DEC' : pd.NA, 'REC': pd.NA },  #20   Nonsense, but should pass
```
mwaddoups, line 290: This looks great! Really nice work.

validator903/validators.py
mwaddoups: The only reason I flag this one is that `pd.to_datetime` doesn't have amazing performance, so from a speed perspective if we can avoid multiple uses of it where possible that would be great.

mwaddoups (changes requested):
Really nice work on this, this is a proper serious piece of work and it's looking great and very readable. This is work to be proud of.

I've noted a couple really minor spots to think about - but very little work needed here.

Also - I noticed from the commits it looks like you might be doing your editing on GitHub? Just wanted to make sure you're aware that you can continue working on the PR from repl.it - just continue committing and pushing on the same branch. 

If your repl.it isn't up to date with the latest changes, you can either make a new replit and check out the branch, or go to an existing one, check out the branch, and click "Pull" to bring down your upstream changes from GitHub.

validator903/validators.py
PhilDeakin: For some reason the 'DEC' field in the episode.csv I was working with ended up in the dd/MM/yy format which is why I added the fillna, but I've just downloaded it again and it's dd/MM/yyyy.. strange, but anyway, I'm relatively sure the CLA system refuses CSV uploads with dates not in dd/MM/yyyy and even if not there are various errors enforcing that dates are in the format dd/MM/yyyy, so I'll remove the fillna. Thanks.

validator903/validators.py
PhilDeakin: Without the DECOM fillna line, 
The " grouped_decom_by_child = df.groupby(['CHILD'])['DECOM'].idxmax(skipna=True) " line throws an error, the skipna parameter doesn't seem to do anything and it's the default anyway. I couldn't think of another way round it. Maybe we should leave it, but mark it with a comment saying "watch for potential future issues" or something like that. 

Having thought about it a bit more, I don't think the DECOM is allowed to be null when you upload it to CLA so it shouldn't matter.

mwaddoups: Okay - yeah, we can leave this for now then I agree and add a comment - I think that's a good idea.

validator903/validators.py
PhilDeakin: As there are hardly any changes this time I've done them here on GitHub, thanks for the info about replit, I didn't realise,  I'll make major changes from there from now on.

mwaddoups (approved):
This is great work and thanks for making those final changes. Nice one to have the first proper advanced rule in the codebase.

PR #306, opened 2021-08-25
113 validator written

### LINKED ISSUE #18: Error code 113 - Date matching child and adopter(s) is not a valid date.
#### Error code
113
#### Description to display
Date matching child and adopter(s) is not a valid date.
If the child has been adopted during the year, the date should be entered in the format DD/MM/YYYY and should contain a valid date.
#### DfE suggested coding
If provided `<DATE_MATCH>` must be in a valid format and contain a valid date  Format should be DD/MM/YYYY.

crossls:
created rule 113

closes #18


validator903/validators.py
```diff
@@ -383,3 +383,27 @@ def _validate(dfs):
...
+        code='113',
+        description='Date matching child and adopter(s) is not a valid date.',
+        affected_fields=['DATE_MATCH'],
+    )
+
+    def _validate(dfs):
+        if 'ad1' not in dfs:
+            return {}
+        else:
+            ad1 = dfs['ad1']
```
mwaddoups: Bit of a hard to spot bug you've ran into here.

There actually isn't a table called `'ad1'` - the key you need is uppercase `'AD1'`. However, I don't have a way of checking this automatically.

So what's happening is that because `'ad1'` is never in `dfs` (except in your test case) this rule would never run in production (since line 395 would always be true).

Either way, you can fix it by changing to AD1 instead, and changing your test as well.

As a reminder, the list of tables is at the top of `config.py` - and case does matter.

mwaddoups (changes requested):
Nice one on putting in the first rule!

Hard-to-spot bug you've ran into here - see the comments but should be an easy fix.

validator903/validators.py
crossls: think this is ok now :)

mwaddoups (approved):
Looks great now - approved!

PR #310, opened 2021-08-25
Added validation for error code 134.

### LINKED ISSUE #29: Error code 134 - Data on adoption should not be entered for the OC3 cohort.
#### Error code
134
#### Description to display
Data on adoption should not be entered for the OC3 cohort.
If returning OC3 data for child, then adoption information must not be completed.
#### DfE suggested coding
If any of `<IN_TOUCH>`; `<ACTIV>` or `<ACCOM>` are provided, then `<DATE_INT>`; `<DATE_MATCH>`; `<FOSTER_CARE>`; `<NB_ADOPTR>`; `<SEX_ADOPTR>` and `<LS_ADOPTR>` must not be provided

WMossDCC:
Added validation to resolve it returning OC3 data for child, then adoption information must not be completed.

Closes #29

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -522,3 +522,42 @@ def validate_113():
             return {'AD1': validation_error_locations.tolist()}
     
     return error, _validate
+
+def validate_134():
+    error = ErrorDefinition(
+        code='134',
+        description='Data on adoption should not be entered for the OC3 cohort.',
+        affected_fields=['IN_TOUCH','ACTIV','ACCOM','DATE_INT','DATE_MATCH','FOSTER_CARE','NB_ADOPTR','SEX_ADOPTR','LS_ADOPTR'],
+    )
+
+    def _validate(dfs):
+        if 'OC3' not in dfs or 'AD1' not in dfs:
+            return {}
+        else:
+            oc3 = dfs['OC3']
+            ad1 = dfs['AD1']
+
+            all_data = oc3.merge(ad1, how='left', on='CHILD')
+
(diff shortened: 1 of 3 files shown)


mwaddoups:
Just noting this is failing test currently - might be worth double checking and fixing that before I review?

validator903/validators.py
```diff
@@ -522,3 +522,42 @@ def _validate(dfs):
...
+    )
+
+    def _validate(dfs):
+        if 'OC3' not in dfs or 'AD1' not in dfs:
+            return {}
+        else:
+            oc3 = dfs['OC3']
+            ad1 = dfs['AD1']
+
+            all_data = oc3.merge(ad1, how='left', on='CHILD')
```
mwaddoups: I think this is likely the culprit for the failing test.

Because you are merging left to OC3, you keep every entry in OC3 and discard entries in AD1. However, that means the merged dataframe is the length of the OC3 return, not the AD1 return.

In your tests this passes because your OC3 and AD1 have the same length. It's probably worth adding an extra line to AD1 that is not present in OC3.

The solution here will be to swap `oc3` and `ad1` (I think) so you have
```
all_data = ad1.merge(oc3, how='left', on='CHILD')
```

If this is a one-to-one merge, `all_data` will be the same length as `ad1` and you'll be able to use the same boolean mask for both.

mwaddoups (approved):
This is great - congrats on figuring out the more advanced stuff too! Don't see any issues with the way you've implemented this, so happy to approve.

PR #312, opened 2021-08-26
Adding errors 142 and 148

### LINKED ISSUE #31: Error code 142 - A new episode has started, but the previous episode has not ended.
#### Error code
142
#### Description to display
A new episode has started, but the previous episode has not ended.
Date and reason the previous episode ceased must be completed, before a new episode can start.  Check the date the last episode ended and complete the previous episode accordingly.
#### DfE suggested coding
For any episode prior to the current episode both `<DEC>` and `<REC>` must be provided

### LINKED ISSUE #37: Error code 148 - Date episode ceased and reason episode ceased must both be coded, or both left blank.
#### Error code
148
#### Description to display
Date episode ceased and reason episode ceased must both be coded, or both left blank.
The date and reason an episode ceased should either both be coded, or both be left blank.  Check which item is missing; if the episode is an open episode, still continuing at the end of the statistical year, both these items must be left blank.
#### DfE suggested coding
If `<DEC>` or `<REC>` is provided, then both `<DEC>` and `<REC>` must contain a value

PhilDeakin:
Closes #31
Closes #37

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -494,7 +494,6 @@ def test_validate_119():
 
     assert result == {'PlacedAdoption': [1, 2]}
 
-"""
 def test_validate_142():
     fake_data = pd.DataFrame([
     { 'CHILD' : '111', 'DECOM' : '01/06/2020', 'DEC' : '04/06/2020', 'REC': 'X1' },  #0
@@ -533,7 +532,7 @@ def test_validate_148():
     { 'CHILD' : '111', 'DECOM' : '01/06/2020', 'DEC' : '04/06/2020', 'REC': 'X1' },  #0  
     { 'CHILD' : '111', 'DECOM' : '05/06/2020', 'DEC' : '06/06/2020', 'REC': 'X1' },  #1
     { 'CHILD' : '111', 'DECOM' : '06/06/2020', 'DEC' : '08/06/2020', 'REC': 'X1' },  #2
-    { 'CHILD' : '111', 'DECOM' : '08/06/2020', 'DEC' : '05/06/2020', 'REC': 'X1' },  #3  
+    { 'CHILD' : '111', 'DECOM' : '08/06/2020', 'DEC' : '05/06/2020', 'REC': 'X1' },  #3   Fails
     { 'CHILD' : '222', 'DECOM' : '05/06/2020', 'DEC' : '06/06/2020', 'REC': pd.NA }, #4   Fails
     { 'CHILD' : '333', 'DECOM' : '06/06/2020', 'DEC' : '07/06/2020', 'REC': 'E11' }, #5  
     { 'CHILD' : '333', 'DECOM' : '07/06/2020', 'DEC' : pd.NA, 'REC': pd.NA },        #6
@@ -559,5 +558,4 @@ def test_validate_148():
 
     result = error_func(fake_dfs)
 
-    assert result == {'Episode': [4,10,14]}
-"""
+    assert result == {'Episode': [3,4,10,14]}
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -493,3 +493,71 @@ def test_validate_119():
     result = error_func(fake_dfs)
 
     assert result == {'PlacedAdoption': [1, 2]}
+
+"""
+def test_validate_142():
+    fake_data = pd.DataFrame([
+    { 'CHILD' : '111', 'DECOM' : '01/06/2020', 'DEC' : '04/06/2020', 'REC': 'X1' },  #0
+    { 'CHILD' : '111', 'DECOM' : '05/06/2020', 'DEC' : '06/06/2020', 'REC': pd.NA },  #1  Fails
+    { 'CHILD' : '111', 'DECOM' : '06/06/2020', 'DEC' : '08/06/2020', 'REC': 'X1' },  #2
+    { 'CHILD' : '111', 'DECOM' : '08/06/2020', 'DEC' : '05/06/2020', 'REC': 'X1' },  #3  
+    { 'CHILD' : '222', 'DECOM' : '05/06/2020', 'DEC' : '06/06/2020', 'REC': pd.NA }, #4
+    { 'CHILD' : '333', 'DECOM' : '06/06/2020', 'DEC' : pd.NA, 'REC': 'E11' },        #5   Fails      
+    { 'CHILD' : '333', 'DECOM' : '07/06/2020', 'DEC' : pd.NA, 'REC': pd.NA },        #6
+    { 'CHILD' : '444', 'DECOM' : '08/06/2020', 'DEC' : '09/06/2020', 'REC': 'X1' },  #7
+    { 'CHILD' : '444', 'DECOM' : '09/06/2020', 'DEC' : '10/06/2020', 'REC': 'E11' }, #8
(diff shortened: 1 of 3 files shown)


PhilDeakin:
These two are very similar to the 388 error but a bit simpler.

validator903/validators.py
```diff
@@ -956,3 +956,66 @@ def _validate(dfs):
...
+    return error, _validate
+
+def validate_148():
+    error = ErrorDefinition(
+        code='148',
+        description='Date episode ceased and reason episode ceased must both be coded, or both left blank.',
+        affected_fields=['DEC','REC'],
+    )
+    
+    def _validate(dfs):
```
mwaddoups, line 997: I think this approach is overcomplex for this rule (but may be missing something) -  to me this looks like a simple rule that we can just do without grouping?

validator903/validators.py
```diff
@@ -956,3 +956,66 @@ def _validate(dfs):
...
+        if 'Episode' not in dfs:
+            return {}
+        else:
+            df = dfs['Episode']
+            df['DECOM'] = pd.to_datetime(df['DECOM'], format='%d/%m/%Y', errors='coerce')
+            df['DEC'] = pd.to_datetime(df['DEC'], format='%d/%m/%Y', errors='coerce')
+
+            df['DECOM'] = df['DECOM'].fillna('01/01/1901') #Watch for potential future issues
+            df = df.sort_values(['CHILD','DECOM'])
```
mwaddoups: I think you can skip the sort here, since `idxmax` will find the maximum DECOM anyway

validator903/validators.py
```diff
@@ -956,3 +956,66 @@ def _validate(dfs):
...
+
+            grouped_decom_by_child = df.groupby(['CHILD'])['DECOM'].idxmax(skipna=True)
+            df['DECOM'] = df['DECOM'].replace('01/01/1901',pd.NA)
+
+            #Dataframe with the maximum DECOM removed
+            df = df.loc[~df.index.isin(grouped_decom_by_child),:]
```
mwaddoups: I would try to avoid this pattern of re-using the same variable name to refer to a different thing as it's easy to introduce bugs. Maybe just call this `ended_episodes_df`.

I think if you also rename `grouped_decom_by_child` to `index_of_last_episodes`, you probably don't need the comment as the code itself should be readable.

tests/test_validators.py
```diff
@@ -493,3 +493,69 @@ def test_validate_119():
...
+    result = error_func(fake_dfs)
+
+    assert result == {'Episode': [1,5,13]}
+
+def test_validate_148():
+    fake_data = pd.DataFrame([
+    { 'CHILD' : '111', 'DECOM' : '01/06/2020', 'DEC' : '04/06/2020', 'REC': 'X1' },  #0  
+    { 'CHILD' : '111', 'DECOM' : '05/06/2020', 'DEC' : '06/06/2020', 'REC': 'X1' },  #1
+    { 'CHILD' : '111', 'DECOM' : '06/06/2020', 'DEC' : '08/06/2020', 'REC': 'X1' },  #2
+    { 'CHILD' : '111', 'DECOM' : '08/06/2020', 'DEC' : '05/06/2020', 'REC': 'X1' },  #3   Fails
```
mwaddoups: I don't think this test case should fail for 148 as both DEC and REC are filled?

mwaddoups (changes requested):
Thanks for putting these in Phil!

142 looks good - I've made some more minor comments as little things to look for, but overall that all looks good.

However I think 148 just needs a simpler approach? I may be missing something here but that looks to me like a simple rule. Keen to hear your thoughts!

validator903/validators.py
PhilDeakin: You're right, I was over/under thinking this. I think I was trying to stop errors 142 & 148 being thrown on the same line, but it doesn't matter.

validator903/validators.py
PhilDeakin: Removed

tests/test_validators.py
PhilDeakin: Corrected

validator903/validators.py
```diff
@@ -956,3 +956,54 @@ def _validate(dfs):
...
+    
+    def _validate(dfs):
+        if 'Episode' not in dfs:
+            return {}
+        else:
+            df = dfs['Episode']
+            df['DECOM'] = pd.to_datetime(df['DECOM'], format='%d/%m/%Y', errors='coerce')
+            df['DEC'] = pd.to_datetime(df['DEC'], format='%d/%m/%Y', errors='coerce')
+
+            mask = (df['CHILD'].notna()) & (df['DECOM'].notna()) & (((df['DEC'].isna()) & (df['REC'].notna())) | ((df['DEC'].notna()) & (df['REC'].isna())))
```
mwaddoups: I'm not sure you need DECOM here at all actually to cover this rule - I think you can just check DEC and REC (and remove line 1002 as well)

tests/test_validators.py
```diff
@@ -493,3 +493,69 @@ def test_validate_119():
...
+    { 'CHILD' : '6666', 'DECOM' : '12/06/2020', 'DEC' : '13/06/2020', 'REC': 'X1' },  #11
+    { 'CHILD' : '6666', 'DECOM' : '13/06/2020', 'DEC' : pd.NA, 'REC': 'X1' },         #12   Fails
+    { 'CHILD' : '6666', 'DECOM' : '14/06/2020', 'DEC' : '15/06/2020', 'REC': 'X1' },  #13
+    { 'CHILD' : '6666', 'DECOM' : '15/06/2020', 'DEC' : '16/06/2020', 'REC': pd.NA },  #14   Fails
+    { 'CHILD' : '77777', 'DECOM' : '16/06/2020', 'DEC' : '17/06/2020', 'REC': 'X1' }, #15
+    { 'CHILD' : '77777', 'DECOM' : '17/06/2020', 'DEC' : '18/06/2020', 'REC': 'X1' }, #16
+    { 'CHILD' : '77777', 'DECOM' : '18/06/2020', 'DEC' : '19/06/2020', 'REC': pd.NA },   #17 Fails
+    { 'CHILD' : '999', 'DECOM' : '31/06/2020', 'DEC' : pd.NA, 'REC': pd.NA },  #18   Nonsense date, but should pass
+    { 'CHILD' : '123', 'DECOM' : pd.NA, 'DEC' : pd.NA, 'REC': 'E11'},  #19   Nonsense dates, but should pass
+    { 'CHILD' : pd.NA, 'DECOM' : pd.NA, 'DEC' : pd.NA, 'REC': pd.NA },  #20   Nonsense, but should pass
```
mwaddoups: This feels like quite a large set of test cases for the rule now given it's quite simple (there's only 4 cases really I think, DEC/REC present or not present). Might be worth using a simpler set of data for that?

However, in general I don't mind having some redundancy in tests - so I'll leave shortening this up to you.

mwaddoups (changes requested):
Thanks for getting back on this so quick! Looking better now - have just noted a few minor ways to make the 148 implementation even simpler.

validator903/validators.py
PhilDeakin: Done

tests/test_validators.py
PhilDeakin: I've shorten the amount of test rows.

mwaddoups (approved):
Looks good now - thanks Phil!

Issue #105, opened 2021-07-29
Error code 184 - Date of decision that a child should be placed for adoption is before the child was born.


mwaddoups:
#### Error code
184
#### Description to display
Date of decision that a child should be placed for adoption is before the child was born.
#### DfE Long-form description of error
If the local authority decides that the child should be placed for adoption the date of this decision cannot be before the child was born.
#### DfE suggested coding
Where provided `<DATE_PLACED>` must be `> = ` `<DOB>` 

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

PR #314, opened 2021-08-27
(with error to look into further) validation rule 184

jfan2006:
Trying to do the rule 184 #105 but still can't figure out how to do it... 
(saying error in line 24 of validators.py)
Could you help me a bit more on this please? Thanks


validator903/validators.py
```diff
@@ -1,6 +1,33 @@
...
+            child_record = dfs['Header']
+            ad1 = dfs['AD1']
+
+            all_data = ad1.reset_index().merge(child_record, how='left', on='CHILD')
+
+            mask = (all_data['DATE_PLACED'] >= all_data['DOB']) | all_data['DATE_PLACED'].isna()
+
+            validation_error = ~mask
+            
+            validation_error_locations = ad1.index[validation_error]['index'].unique()
```
mwaddoups: I think you just don't need the `ad1.index` part - the `'index'` column is just a regular old column, so 

```
ad1[validation_error]['index'].unique()
```

should do the trick!

mwaddoups:
I added a comment on line 24 - take a look and see if that helps.

validator903/validators.py
jfan2006: I have just tried as you suggested but it still has errors.

mwaddoups: Are you able to provide some more details - like screenshots of the error or something? That might be helpful in diagnosing.

dezog: Hi Fan, are you sure the DataFrame has a column called `'index'`? 
Maybe try keeping the `.index` and dropping the `['index']`:
```
ad1.index[validation_error].unique()
```

dezog: Looks like `all_data` will have a column called `'index'` due to  the `reset_index()` in line 18, but `ad1` won't have had its index reset. So maybe instead try:
```
ad1.index[validation_error].unique()
```
That make sense, @mwaddoups?

mwaddoups: Now I'm confused! I think it needs to be `all_data` to read the appropriate index values off.

```
all_data[validation_error]['index'].unique()
```

dezog (changes requested):
Just setting this to show 'changes requested'. 

Let me know if the edit Mark suggested in https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/pull/314#discussion_r700194448 doesn't do the trick, but it looks like it should do. 🤞

jfan2006:
Tried and made changes based on Mark's suggestion above, and also dates comparison. 
But this still has errors....

tests/test_validators.py
```diff
@@ -1,6 +1,23 @@
...
+    fake_data_header = pd.DataFrame({
+        'CHILD': ['111', '112', '113', '114'], 
+        'DOB': ['01/10/2017', '31/05/2018', '10/03/2019', '19/08/2020'],
+    })
+    fake_data_ad1 = pd.DataFrame({
+        'CHILD': ['111', '112', '113', '114'], 
+        'DATE_PLACED': ['01/10/2017', '10/02/2019', '10/02/2019', pd.NA],
+    })
+
+    fake_dfs = {'Header': fake_data_header, 'AD1': fake_data_ad1}
```
dezog: AD1 > PlacedAdoption

tests/test_validators.py
```diff
@@ -1,6 +1,23 @@
...
+    
+    error_defn, error_func = validate_184()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'AD1': [2]}
```
dezog: AD1 > PlacedAdoption

validator903/validators.py
```diff
@@ -1,6 +1,36 @@
...
 
+def validate_184():
+    error = ErrorDefinition(
+        code='184',
+        description='Date of decision that a child should be placed for adoption is before the child was born.',
+        affected_fields=['DATE_PLACED','DOB'],
+    )
+
+    def _validate(dfs):
+        if 'Header' not in dfs or 'AD1' not in dfs:
```
dezog: AD1 > PlacedAdoption

validator903/validators.py
```diff
@@ -1,6 +1,36 @@
...
+            all_data['DATE_PLACED'] = pd.to_datetime(all_data['DATE_PLACED'],format='%d/%m/%Y',errors='coerce')
+            all_data['DOB'] = pd.to_datetime(all_data['DOB'],format='%d/%m/%Y',errors='coerce')
+
+            mask = (all_data['DATE_PLACED'] >= all_data['DOB']) | all_data['DATE_PLACED'].isna()
+
+            validation_error = ~mask
+            
+            validation_error_locations = all_data[validation_error]['index'].unique()
+        
+            return {'AD1': validation_error_locations.tolist()}
```
dezog: AD1 > PlacedAdoption

validator903/validators.py
```diff
@@ -1,6 +1,36 @@
...
+        description='Date of decision that a child should be placed for adoption is before the child was born.',
+        affected_fields=['DATE_PLACED','DOB'],
+    )
+
+    def _validate(dfs):
+        if 'Header' not in dfs or 'AD1' not in dfs:
+            return {}
+        else:
+            child_record = dfs['Header']
+            ad1 = dfs['AD1']
```
dezog: AD1 > PlacedAdoption

dezog (changes requested):
Hey Fan, think I've figured this one out - it's looking for DATE_PLACED in the AD1 table, but that's actually a field of the PlacedAdoption table. 

That's also causing an error in one of the integration tests where merging AD1 & Header leads to 'DOB_x' & 'DOB_y' columns since they both have that field, so accessing DOB raises a KeyError - but switching to PlacedAdoption should fix this.

I've tried to indicate all the relevant lines to change in the code, hopefully that does it!

dezog (approved):

PR #317, opened 2021-08-27
Added errors 214 & 222, so I can delete the very first push request I…

### LINKED ISSUE #119: Error code 214 - Placement location information not required.
#### Error code
214
#### Description to display
Placement location information not required.
Placement location information is not required for children with a legal status of V3 or V4.
#### DfE suggested coding
If `<LS>` = 'V3' or 'V4' then `<PL_POST>` and `<URN>` should not be provided

### LINKED ISSUE #155: Error code 366 - A child cannot change placement during the course of an individual short- term respite break.
#### Error code
366
#### Description to display
A child cannot change placement during the course of an individual short- term respite break.
Where the legal status is V3 (meaning individual short-term breaks under a respite agreement are being recorded), the reason for new episode must be S.  Check status of child to ensure that it is a genuine respite case. If
#### DfE suggested coding
Where `<LS>` = 'V3' then `<RNE>` must be 'S'

### LINKED ISSUE #297: Error code 222 - Ofsted Unique reference number (URN) should not be recorded for this placement type.
#### Error code
222
#### Description to display
Ofsted Unique reference number (URN) should not be recorded for this placement type.
Ofsted URN should not be recorded for the following placement types:
- H5 – semi-independent living accommodation not subject to Children’s Homes Regulations
= P1 - Placed with own parent(s) or other person with parental responsibility
- P2 - Independent living, for example, in flat, lodgings, bedsit, bed and breakfast (B & B) or with friends, with or without formal support
- P3 - Residential employment
- R1 - Residential care home
- R2 – National Health Service (NHS) / health trust or other establishment providing medical or nursing care
- R5 - Young offender institution or prison
- T0 - All types of temporary move
- T1 - Temporary periods in hospital
- T2 -Temporary absences of the child on holiday
- T3 - Temporary accommodation whilst normal foster carer(s) is/are on holiday
#### DfE suggested coding
If `<URN>` provided and `<URN>` not = ‘XXXXXX’, and where `<PL>` = ‘H5’; ‘P1’ ‘P2’ ‘P3’; ‘R1’; ‘R2’; ‘R5’; ‘T0’ ‘T1’; ‘T2’; ‘T3’; ‘T4’ or Z1 then `<URN>` should not be provided

PhilDeakin:
… sent.

Closes #119 
Closes #297 
Closes #155

### DIFF
--- a/validator903/config.py
+++ b/validator903/config.py
@@ -67,4 +67,6 @@ configured_errors = [
     validate_142(),
     validate_148(),
     validate_366(),
+    validate_222(),
+    validate_214(),
 ]
--- a/validator903/config.py
+++ b/validator903/config.py
@@ -66,4 +66,5 @@ configured_errors = [
     validate_1009(),
     validate_142(),
     validate_148(),
+    validate_366(),
 ]
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -1038,12 +1038,12 @@ def validate_214():
     )
 
     def _validate(dfs):
-        if 'Episode' not in dfs:
+        if 'Episodes' not in dfs:
             return {}
         else:
-            df = dfs['Episode']
+            df = dfs['Episodes']
             mask = df['LS'].isin(['V3','V4']) & ((df['PL_POST'].notna()) | (df['URN'].notna()))
-            return {'Episode': df.index[mask].tolist()}
+            return {'Episodes': df.index[mask].tolist()}
 
     return error, _validate
 
@@ -1055,13 +1055,13 @@ def validate_222():
     )
 
     def _validate(dfs):
-        if 'Episode' not in dfs:
+        if 'Episodes' not in dfs:
             return {}
         else:
-            df = dfs['Episode']
+            df = dfs['Episodes']
             place_code_list = ['H5','P1','P2','P3','R1','R2','R5','T0','T1','T2','T3','T4','Z1']
             mask = (df['PLACE'].isin(place_code_list)) & (df['URN'].notna()) & (df['URN'] != 'XXXXXX')
-            return {'Episode': df.index[mask].tolist()}
+            return {'Episodes': df.index[mask].tolist()}
 
     return error, _validate
 
@@ -1073,11 +1073,11 @@ def validate_366():
     )
 
     def _validate(dfs):
-        if 'Episode' not in dfs:
+        if 'Episodes' not in dfs:
             return {}
         else:
-            df = dfs['Episode']
+            df = dfs['Episodes']
             mask = (df['LS'] == 'V3') & (df['RNE'] != 'S')
-            return {'Episode': df.index[mask].tolist()}
+            return {'Episodes': df.index[mask].tolist()}
 
     return error, _validate
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -1027,7 +1027,7 @@ def validate_148():
     return error, _validate
 
 
-def validate_214 ():
+def validate_214():
     error = ErrorDefinition(
         code='214',
         description='Placement location information not required.',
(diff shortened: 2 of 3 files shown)


validator903/config.py
```diff
@@ -66,4 +66,5 @@
     validate_1009(),
     validate_142(),
     validate_148(),
+    validate_366(),
```
mwaddoups, line 69: You need to add the other two rules here I think (and double check the integration tests run).

validator903/validators.py
```diff
@@ -1025,3 +1025,56 @@ def _validate(dfs):
...
+
+    def _validate(dfs):
+        if 'Episode' not in dfs:
+            return {}
+        else:
+            df = dfs['Episode']
+            mask = df['LS'].isin(['V3','V4']) & ((df['PL_POST'].notna()) | (df['URN'].notna()))
+            return {'Episode': df.index[mask].tolist()}
+
+    return error, _validate
```
mwaddoups, line 1048: A question I don't know the answer to - does URN 'XXXXX' also count as not providing a URN? Or should that fail validation here?

mwaddoups: Based on 222's description I wonder if XXXXXXX should be maybe allowed.

validator903/validators.py
```diff
@@ -1025,3 +1025,56 @@ def _validate(dfs):
...
+
+def validate_222():
+    error = ErrorDefinition(
+        code='222',
+        description='Ofsted Unique reference number (URN) should not be recorded for this placement type.',
+        affected_fields=['URN'],
+    )
+
+    def _validate(dfs):
+        if 'Episode' not in dfs:
```
mwaddoups: Just noticed a rather large bug - this actually affects some of the preexisting rules as well.

The table is `df['Episodes']`, not `df['Episode']` - so any rule which is written against `'Episode'` doesn't run!

I'll correct the existing rules in a separate PR and try to write a test to catch that behaviour going forward.

mwaddoups (changes requested):
Logic looks good here and nice code as usual.

Just noticed a fairly major bug affecting both these and other rules that isn't tested for properly - I'm going to fix that outside this rule. Otherwise had a few minor questions about the specifics of 214 - not sure if you'll know the answer?

validator903/validators.py
PhilDeakin: We enter XXXXXX when the error demands an URN, but the placement they're at doesn't have one because it's inspected by the CQC rather than OfSted or the placement's outside England (and probably a few other reasons I can't think of off the top of my head), in terms of this particular error, the guidance doc doesn't mention XXXXXX and it does in various other errors. I think it's okay to leave it as it is, particularly since the LS codes of V3 and V4 hardly ever occur, so I don't think this error is likely to ever get thrown and I'd assume that's the same for other councils.

validator903/config.py
PhilDeakin: Added and ran pytest

validator903/validators.py
PhilDeakin: I've left it as 'Episode'. But, I'll switch over when all the codes been amended to 'Episodes'

mwaddoups: @PhilDeakin I've fixed this (in #320) and brought your branch up to date with main - you can see the tests are now failing, but if you change to Episodes everywhere they should pass.

You will need to pull changes in from Github in replit (to incorporate the merge), then make the changes and push back

validator903/validators.py
mwaddoups: That's very helpful, agree then we can leave it. Thanks Phil!

validator903/validators.py
PhilDeakin: Pushed back, thanks.

PhilDeakin: @mwaddoups done

mwaddoups (approved):
All good now - thanks for the changes and good to see these in! I'll close the other PR as well.

PR #321, opened 2021-08-31
Fixed the wrong table name and added tests going forward

### LINKED ISSUE #320: Validators which look for the wrong table pass tests silently
If a validator has the table name spelt wrong (in both validator and tests), it will silently pass tests. I need to add tests to try and detect this.

mwaddoups:
Added the tests for wrong table name, and updated the rules where that had been entered incorrectly.

Closes #320


dezog (approved):
looks good!

PR #322, opened 2021-08-31
Added error 628

### LINKED ISSUE #271: Error code 628 - Motherhood details are not required for care leavers who have not been looked after during the year.
#### Error code
628
#### Description to display
Motherhood details are not required for care leavers who have not been looked after during the year.
Mother status field should be left blank for young people for whom care leaver information is returned.  However, if the care leaver also had episode information recorded for this collection year, motherhood status should be returned.
#### DfE suggested coding
If no episode record for current year and any of `<IN_TOUCH>`, `<ACTIV>` or `<ACCOM>` have been provided then `<MOTHER>` should not be provided

PhilDeakin:
Added 628

Closes #271


PhilDeakin:
Clears issue 271

validator903/validators.py
```diff
@@ -1080,4 +1080,30 @@ def _validate(dfs):
...
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs or 'Header' not in dfs or 'OC3' not in dfs:
+            return {}
+        else:
+            hea = dfs['Header']
+            epi = dfs['Episodes']
+            oc3 = dfs['OC3']
+            
+            oc3_has_in_touch_activ_or_accom = oc3[(oc3['IN_TOUCH'].notna()) | (oc3['ACTIV'].notna()) | (oc3['ACCOM'].notna())]
```
mwaddoups: A good trick for doing the mask in less space is
```
oc3[['IN_TOUCH', 'ACTIV', 'ACCOM']].notna().any(axis=1)
```

This
- `oc3[['IN_TOUCH', 'ACTIV', 'ACCOM']]` Selects just the appropriate columns.
- `.notna()` - returns a dataframe with the same columns and True/False values for nona
- `.any(axis=1)` - If any values are True, returns True. The `axis=1` specifies to work across rows rather than columns. This therefore aggregates the DataFrame to a Series.

validator903/validators.py
```diff
@@ -1080,4 +1080,30 @@ def _validate(dfs):
...
+
+            child_in_header_not_in_episode_file = ~hea['CHILD'].isin(epi['CHILD'])
+            child_in_header_in_oc3_with_appropriate_values = hea['CHILD'].isin(oc3_has_in_touch_activ_or_accom['CHILD'])
```
mwaddoups: This one is tagged as an advanced rule - that tends to mean I'm looking for a merging or grouping solution. The video for this is coming out this morning so I would suggest looking through those materials before coming back to this.

Your solution using `.isin` will work, but will likely be orders of magnitude slower on the performance side - this is why merging tends to be preferred whenever we need to look across multiple tables.

In terms of approach, this is again one where the solution is complicated (you're good at picking these!). I think a good approach would be
- Inner merge oc3 and episodes on CHILD (this should be empty if there's no errors)
- Inner merge header with index column, (this will make sense once you watch the videos) with that output, which should give a list of error candidates.

mwaddoups: Actually - I will say that for smaller datasets (~1000-2000 children) the `isin` approach isn't much worse on performance, having just checked it. The merge will perform better as the datasets get larger.

I'd still suggest going for a merge with this anyway - the solution ends up being more readable, and it's a good opportunity to learn more about that side for other work you might do in Python!

mwaddoups (changes requested):
Similar to another rule we looked at I think - your solution works, but as this is an advanced rule a more readable solution exists using merging. In general if I've tagged something as advanced it's because I think it will need a merge or group by.

Happy for you to have a crack at this (by reading https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#database-style-dataframe-or-named-series-joining-merging first) - but ideally I think watching the new videos that should be coming out this morning should give a good primer.

Good work on getting to a workable solution with a bit more of a limited toolset though!

validator903/validators.py
PhilDeakin: Thanks that's cool.

validator903/validators.py
PhilDeakin: I've changed this to merges... I played about with it, but I couldn't get the two the inner joins in your suggestion to work though, so it's a left join then inner join.

validator903/validators.py
```diff
@@ -1080,4 +1080,32 @@ def _validate(dfs):
...
+            hea = hea.reset_index()
+            oc3_no_nulls = oc3[oc3[['IN_TOUCH','ACTIV','ACCOM']].notna().any(axis=1)]
+
+            hea_merge_epi = hea.merge(epi,how='left',on='CHILD',indicator=True)
+            hea_not_in_epi = hea_merge_epi[hea_merge_epi['_merge'] == 'left_only']
+
+            cohort_to_check = hea_not_in_epi.merge(oc3_no_nulls,how='inner',on='CHILD')
+            error_cohort = cohort_to_check[cohort_to_check['MOTHER'].notna()]
+
+            return {'Header': error_cohort['index'].to_list()}
```
mwaddoups: I think you might get away with it here, but good practice is to get only the unique values from the index. Since you are merging header to episodes, the header index may get repeated.

In this particular case, I think OC3 and header will only have one entry per child but am not 100% sure, and because we only select children with no match they will only appear once.

However I'd suggest doing this anyway - it's a good habit to get into and avoids someone having to make the leap of faith. You can either `list(set(error_cohort['index']))` or `error_cohort['index'].unique().tolist()` I think.

mwaddoups (changes requested):
Nice one Phil, I like the solution you've used here and hopefully learning more about the merging was interesting too!

Fairly minor point here (and again more of a learning point than a functionality point) - but after that small change should be approved.

validator903/validators.py
PhilDeakin: Done, I've sorted it as well, probably not needed, but I doubt it's an expensive operation.

mwaddoups: Agree - it's not expensive. Cool - this looks good!

mwaddoups (approved):
Thanks for all the changes Phil, approved and in

PR #323, opened 2021-08-31
Added validation rule 621.

### LINKED ISSUE #267: Error code 621 - Mother's field has been completed but date of birth shows that the mother is younger than her child.
#### Error code
621
#### Description to display
Mother's field has been completed but date of birth shows that the mother is younger than her child.
Where the mother's child date of birth is completed it should be greater than the mother's date of birth.  Check child's date of birth and mother's date of birth are correct.
#### DfE suggested coding
If provided `<MC_DOB>` should be >` `<DOB>`

jfan2006:
Added validation rule 621.

Closes #267


validator903/validators.py
```diff
@@ -1,6 +1,28 @@
...
+        affected_fields=['DOB', 'MC_DOB'],
+    )
+
+    def _validate(dfs):
+        if 'Header' not in dfs:
+            return {}
+        else:
+            header = dfs['Header']
+           
+            mask = (header['MC_DOB'] > header['DOB']) | header['MC_DOB'].isna()
```
mwaddoups, line 20: Kind of a fun bug here - you need to convert these both to datetime first! Pandas allow this since comparison is defined for strings (e.g. `'b' > 'a') but that won't work for datetimes.

In general with dates you need to convert them, but also make sure you have the correct date format (d/M) and that you handle cases where a date is entered incorrectly without crashing (`errors='coerce'`). Otherwise use `pd.to_datetime` to make the conversion first. 

There should be examples of this in other errors you can look at.

tests/test_validators.py
```diff
@@ -1,6 +1,20 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_621():
+    fake_data = pd.DataFrame({
+        'DOB': ['01/12/2021', '19/02/2016', '31/01/19',  '31/01/19'],
+        'MC_DOB': ['01/01/2021', '19/12/2016',  '31/01/19', pd.NA],
```
mwaddoups: Might be worth adding a test case that checks the string conversion stuff - for example, '31/01/19' vs. '01/02/19' - the second one is a later date, but will show as less than the first in strings.

mwaddoups (changes requested):
Thanks for submitting this Fan.

Small bug here in converting to datetime - and you get a bit lucky (unlucky?) with your test cases that you don't spot it!

dezog (approved):
looks perfect, merging this

dezog (approved):

PR #324, opened 2021-08-31
Added validation rule 612.

### LINKED ISSUE #265: Error code 612 (amended) - Date of birth field has been completed but mother field indicates child is not a mother.
#### Error code
612 (amended)
#### Description to display
Date of birth field has been completed but mother field indicates child is not a mother.
Mother field must be completed if the child is a mother.
#### DfE suggested coding
If `<MOTHER>` = '0' or Null and `<SEX>` = 2 then `<MC_DOB>` should not be provided.

jfan2006:
Added validation rule 612.

Closes #265


validator903/validators.py
```diff
@@ -1,6 +1,28 @@
...
+        affected_fields=['SEX', 'MOTHER', 'MC_DOB'],
+    )
+
+    def _validate(dfs):
+        if 'Header' not in dfs:
+            return {}
+        else:
+            header = dfs['Header']
+
+            mask = (header['MOTHER'].astype(str).isin(['0']) | header['MOTHER'].isna()) & header['SEX'].astype(str).isin(['2']) & header['MC_DOB'].isna()
```
dezog: A very minor style point - when there's only one value to check against, like `header['MOTHER'].astype(str).isin(['0']) `, I think it'd be a little more readable to just say `header['MOTHER'].astype(str) == '0'`

tests/test_validators.py
```diff
@@ -1,6 +1,21 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_612():
+    fake_data = pd.DataFrame({
+        'SEX': [2, '2', '2', 2], 
+        'MOTHER': ['0', '0', pd.NA, pd.NA],
```
dezog: One more test case would be good, just to ensure the error's not triggering when the child is in fact a mother. ie. when MOTHER = 1 and MC_DOB is valid

dezog (changes requested):
Hi Fan,

This is looking to great, only a couple of small things that might be worth tweaking which I've noted as comments.

Other than that though, this is looking good to go! 👍

jfan2006:
Hi @dezog made some changes but there are errors that I can't figure out.... could you help? thanks

dezog:
> 
> 
> Hi @dezog made some changes but there are errors that I can't figure out.... could you help? thanks

Hi Fan, sure thing - I'll give these all a proper look over tomorrow morning. 

At first glance I think this one might be down to needing extra brackets on the `mask =...` line

validator903/validators.py
```diff
@@ -1,6 +1,28 @@
...
+        affected_fields=['SEX', 'MOTHER', 'MC_DOB'],
+    )
+
+    def _validate(dfs):
+        if 'Header' not in dfs:
+            return {}
+        else:
+            header = dfs['Header']
+
+            mask = (header['MOTHER'].astype(str) == '0' | header['MOTHER'].isna()) & header['SEX'].astype(str) == '2' & header['MC_DOB'].isna()
```
dezog: 1. If I'm not mistaken, **`header['MC_DOB'].isna()` should be `header['MC_DOB'].notna()`, and the resulting mask should not then be negated as in line 19**, as this is then the correct code for the errors, and you only need to do that when you initially obtain the valid rows.
2. **This line needs brackets around `header['MOTHER'].astype(str) == '0'`, and around `header['SEX'].astype(str) == '2'`.**

    - This is because 'bitwise operators' like `|` and `&` happen before normal comparisons like `==`, so for example Python ends up trying to do `'0' | header['MOTHER'].isna()` first, which isn't what you want, and throws an error since '0' is a string not a bool (the error message `TypeError: Cannot perform 'or_' with a dtyped [bool] array and scalar of type [bool]` is kind of misleading...)

dezog (changes requested):
Hey Fan, sorry for the delay. As I thought, the errors are being caused by one of those annoying syntax issues (stemming from my earlier suggestion - sorry about that!). I think I've also noticed a small bug in the validation logic.

dezog (approved):

PR #327, opened 2021-09-02
Added error 355

### LINKED ISSUE #146: Error code 355 - Episode appears to have lasted for less than 24 hours.
#### Error code
355
#### Description to display
Episode appears to have lasted for less than 24 hours.
Episodes cannot start and cease on the same day.  An episode which started and ended on the same day should not be shown on the data collection record.
#### DfE suggested coding
`<DECOM>` and `<DEC>` cannot be the same date

PhilDeakin:
Rules 141 & 147 enforce that DECOM and DEC are in the format dd/MM/yyyy so I think we're fine testing as a string.
astype(str) probably isn't needed, but won't matter.


dezog (changes requested):
Hey Phil, 

Nice work, the validation looks good - as you say, converting the dates to datetime wouldn't make a difference in this case. 

For completeness I would add one more test case where the DEC is before the DECOM, to ensure this error can't trigger alongside rule #356. Shouldn't make any practial difference, but it's good to be sure, just in case we make changes to the code in future.

PhilDeakin:
@dezog Done

dezog (approved):
Looks good. Nice one, Phil 👍

dezog:
Closes #327

PR #330, opened 2021-09-07
Added error 630

### LINKED ISSUE #272: Error code 630 - Information on previous permanence option should be returned.
#### Error code
630
#### Description to display
Information on previous permanence option should be returned.
A new episode has started in the latest reporting year, with the reason for new episode being S, therefore data for items relating to previous permanence option must be returned.  Data items:  Previous permanence option, Local authority code where previous permanence option was arranged (unless previous permanence option is Z1), Date of order (unless previous permanence option is Z1),  must be returned for all children starting to be looked after 1 April of the current collection year, this includes children with legal status of V3 or V4.
#### DfE suggested coding
If `<RNE>` = 'S' then `<PREV_PERM>` must be provided as should `<LA_PERM>` and `<DATE_PERM>` if `<PREV_PERM>` is not = 'Z1'

PhilDeakin:
Added error 630

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -784,3 +784,21 @@ def test_validate_586():
     error_defn, error_func = validate_586()
 
     assert error_func(fake_dfs) == {'Missing': [2, 4]}
+
+def test_validate_630():
+    fake_epi = pd.DataFrame({
+        'CHILD': ['0', '1', '2', '3', '4', '5' ,'6', '7' ,'8'],
+        'RNE': [  'T', 'S', 'S', 'P', 'S', 'S', 'S', 'S', 'S'],
+    })
+    fake_pre = pd.DataFrame({
+        'CHILD':      ['2',   '4', '5' ,          '6',  '7' , '8'],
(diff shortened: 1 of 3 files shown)


dezog (changes requested):
Great job on this, it's a pretty complex rule. 

The only things that stand out to me are:

1.  On more complicated rules like this I think it would be good to include a few comments in the `_validate()` function so the thought process behind it is explicit -this should make it easier if we find we need to tweak anything down the line, and also help me review the code a bit faster too

2. This part of the description:

>[previous permanence fields] must be returned for all children starting to be looked after 1 April of the current collection year, this includes children with legal status of V3 or V4.  

This seems to imply that if an `RNE == 'S'` episode started before the current collection year, they don't need to be included in the PreviousPermanence table (though weirdly it's not mentioned in the 'Coding Description'). Does that sound right to you? 

image

PhilDeakin:
Think you're right about it needing to look at RNE of 'S' in this financial year only, I've amended it to do that.

dezog (approved):
Good stuff - approved

Great work this week, have a nice weekend

dezog:
closes #272

PR #333, opened 2021-09-08
Added error 501

### LINKED ISSUE #206: Error code 501 - A new episode has started before the end date of the previous episode.
#### Error code
501
#### Description to display
A new episode has started before the end date of the previous episode.
The date the previous episode ceased must be the same (for a continuation episode) or before the date the current episode commenced.  Check the date the last episode ended and the date the present episode started.
#### DfE suggested coding
`<DECOM>` of current episode must be >= `<DEC>` of previous episode

PhilDeakin:
As above

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -1337,7 +1337,7 @@ def validate_501():
     error = ErrorDefinition(
         code='501',
         description='A new episode has started before the end date of the previous episode.',
-        affected_fields=['DECOM'],
+        affected_fields=['DECOM','DEC'],
     )
 
     def _validate(dfs):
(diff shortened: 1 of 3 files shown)


validator903/validators.py
```diff
@@ -1331,4 +1331,34 @@ def _validate(dfs):
             error_mask = df['MIS_START'].notna() & (df['MIS_START'] <= df['DOB'])
             return {'Missing': df.index[error_mask].to_list()}
 
+    return error, _validate
+
+def validate_501():
+    error = ErrorDefinition(
+        code='501',
+        description='A new episode has started before the end date of the previous episode.',
+        affected_fields=['DECOM'],
```
dezog: Looks like `'DEC'` should also be in the affected fields

validator903/validators.py
```diff
@@ -1331,4 +1331,34 @@ def _validate(dfs):
...
+            epi = epi.reset_index()
+            epi['DECOM'] = pd.to_datetime(epi['DECOM'],format='%d/%m/%Y',errors='coerce')
+            epi['DEC'] = pd.to_datetime(epi['DEC'],format='%d/%m/%Y',errors='coerce')
+
+            epi = epi.sort_values(['CHILD','DECOM'])
+
+            epi_lead = epi.shift(1)
+            epi_lead = epi_lead.reset_index()
+
+            m_epi = epi.merge(epi_lead,left_on='index',right_on='level_0')
```
dezog: Just to make this part nice and explicit, you could include the additional argument: `suffixes=('', '_prev')` when calling merge.

(Then instead of `CHILD_x` and `CHILD_y`, you'd have `CHILD` and `CHILD_prev`)

dezog (changes requested):
Good work Phil, logic looks solid. 

Couple of minor edits in comments

validator903/validators.py
PhilDeakin: Done

validator903/validators.py
PhilDeakin: Done, but are you sure? the potentially incorrect DEC will be on the proceeding line, so won't match up with the error index list we're returning

validator903/validators.py
PhilDeakin: @dezog Done, thanks.

validator903/validators.py
dezog: No you're completely right, my mistake! 

I'll change `affected_fields` back and merge this branch now

dezog (commented):
Merged

PR #335, opened 2021-09-15
SL - improving my code

SLornieCYC:
* Fixed indents on error 356 (my first rule)
* improved merge logic on error NoE (second rule) as found a better way to do it!

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -18,14 +18,13 @@ def validate_NoE():
             episodes_last['DECOM'] = pd.to_datetime(episodes_last['DECOM'],format='%d/%m/%Y',errors='coerce')
             collection_start = pd.to_datetime(dfs['metadata']['collection_start'],format='%d/%m/%Y',errors='coerce')
 
-            episodes['index_copy'] = episodes.index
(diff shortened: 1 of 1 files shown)


validator903/validators.py
```diff
@@ -18,14 +18,13 @@ def _validate(dfs):
             episodes_last['DECOM'] = pd.to_datetime(episodes_last['DECOM'],format='%d/%m/%Y',errors='coerce')
             collection_start = pd.to_datetime(dfs['metadata']['collection_start'],format='%d/%m/%Y',errors='coerce')
 
-            episodes['index_copy'] = episodes.index
             episodes_before_year = episodes[episodes['DECOM'] < collection_start]
 
-            episodes_merged = pd.merge(episodes_before_year, episodes_last, how='left', on=['CHILD'], indicator=True)
+            episodes_merged = episodes_before_year.reset_index().merge(episodes_last, how='left', on=['CHILD'], indicator=True).set_index('index')
```
dezog, line 23: Totally optional, but it's good to have a comment explaining the gist of multi-step lines like this

dezog (approved):
Nice, this makes the code cleaner & clearer. Approved

Issue #107, opened 2021-07-29
Error code 202 - The gender code conflicts with the gender already recorded for this child.


mwaddoups:
#### Error code
202
#### Description to display
The gender code conflicts with the gender already recorded for this child.
#### DfE Long-form description of error
DfE holds an existing record for this child, but the gender details do not match.  If you are certain that these details are correct and previous details are wrong, please amend through the master change function or contact DfE about correcting earlier years.
#### DfE suggested coding
`<SEX>` from current year's record must = `<SEX>` from previous year's record

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

SLornieCYC:
Duplicate of #74

PR #336, opened 2021-09-17
New validator for error code 202

### LINKED ISSUE #74: Error code 202 - The gender code conflicts with the gender already recorded for this child.
#### Error code
202
#### Description to display
The gender code conflicts with the gender already recorded for this child.
DfE holds an existing record for this child, but the gender details do not match.  If you are certain that these details are correct and previous details are wrong, please amend through the master change function or contact DfE about correcting earlier years.
#### DfE suggested coding
`<SEX>` from current year's record must = `<SEX>` from previous year's record

SLornieCYC:
Closes #74 and #107 (duplicated).


tests/test_validators.py
```diff
@@ -1,6 +1,25 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_202():
+    fake_data = pd.DataFrame({
+        'CHILD':['101','102','103','104'],
```
dezog: I'd add a couple of extra cases here: 
- a couple of children each to `fake_data` & `fake_data_prev` who aren't present in the other
- and maybe a case or two where `SEX` is missing or invalid, to make sure that it works as expected - ie. triggering this error only if the child is in both years, with a valid value this year which was not the same last year. 

(as I understand it, if this year's value is invalid that should trigger **Error 101: Gender code is not valid**, but not this one. Do you agree or do you think it would make more sense for a user to have both flagged up?)

validator903/validators.py
```diff
@@ -1,6 +1,32 @@
...
+    def _validate(dfs):
+        if 'Header' not in dfs or 'Header_last' not in dfs:
+            return {}
+        else:
+            header = dfs['Header']
+            header_last = dfs['Header_last']
+
+            header_merged = header.reset_index().merge(header_last, how='left', on=['CHILD'], suffixes=('', '_last'), indicator=True).set_index('index')
+
+            header_matched = header_merged[header_merged['_merge'] == 'both']
```
dezog: Instead of selecting only the matched rows here, you could just take their locations, eg. `in_both_years = header_merged['_merge'] == 'both'`, then do `error_mask = in_both_years & sex_is_different` (perhaps also with `& sex_is_valid`, depending what you think of my comment on the test function )

dezog (changes requested):
Hey Stephen, 

Nice work, this is almost there - I've spotted a subtle bug that's not caught by the test though: 

- `error_mask` is derived from `header_matched`, which is all the children present in both this year's and last year's header files. 
- This means if there are any children present in this year's return but not last year's, then `error_locations = header.index[error_mask]` will throw an IndexError since `header` and `error_mask` are different lengths
- The list of child id's is the same in the two test header tables, so currently this doesn't happen and the test passes

I've added comments with some ideas for fixing it, let me know if anything doesn't make sense

SLornieCYC:
https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/pull/336#pullrequestreview-757516933

~~Can you explain this index issue a bit more @dezog ?  I've only got one header file in this rule (and no `header_matched`) so no merging involved.  From testing the frame filtering it does look to preserve the original index throughout which should be fine for the final `error_locations`?~~

EDIT: Nevermind! I've been looking at the wrong test! Will re-read your comments in context of the right python and get back to you.. 

EDIT2: Now understood and fixed. Thanks! :)

tests/test_validators.py
SLornieCYC: ~~For invalid or missing SEX value I think I'd rather not trigger this error, because that wouldn't make sense with the returned message and highlighted cells here ("Child is looked after but mother field is not completed." and "MOTHER").  Doubling or trebling up on reported errors when invalid data hits other rules just makes it harder to figure out what the underlying problems are that need fixing.~~

~~I'll look to put a few more test cases in for various combinations though.~~

EDIT: Nevermind! I've been looking at the wrong test! Will re-read your comments in context of the right python and get back to you..

SLornieCYC: Okay @dezog added a new commit with corrections and additional tests.  

On reflection I think it is fine for this error to trigger alongside 101 on invalid or missing `SEX` where this is different to the previous year value because that provides additional info to the end user - that there _was_ a value recorded in the previous year (helpful for DQ checking).  

I haven't excluded invalid values in either year so technically there is an edge case where the rule will trigger if a child exists in both years with `SEX` recorded this year but the value is missing or invalid last year - however I think this combination is unlikely to occur as that error should already have been spotted and corrected last year!

dezog: You're right, this isn't like some cases where the additional errors are irrelevant if the value's missing, having that extra bit of info makes good sense from the user's perspective. 

Latest commit is on point, I'll merge this into main now. Cheers Stephen!

dezog (approved):
Requested changes have been added

Issue #40, opened 2021-07-29
Error code 153 - All data items relating to a child's activity or accommodation after leaving care must be coded or left blank.


mwaddoups:
#### Error code
153
#### Description to display
All data items relating to a child's activity or accommodation after leaving care must be coded or left blank.
#### DfE Long-form description of error
Data items relating to activity or accommodation after leaving care should either be coded (all of them) or left blank (all of them).  Enter missing data.
#### DfE suggested coding
Where any of `<IN_TOUCH>`; `<ACTIV>` or `<ACCOM>` is provided, all must be provided

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

Issue #84, opened 2021-07-29
Error code 197 - Reason for no Strengths and Difficulties (SDQ) score is not required if Strengths and Difficulties Questionnaire score is filled in.


mwaddoups:
#### Error code
197
#### Description to display
Reason for no Strengths and Difficulties (SDQ) score is not required if Strengths and Difficulties Questionnaire score is filled in.
#### DfE Long-form description of error
If a child has been looked after for 12 months continuously at 31 March of the current collection year and they were aged between 4 and 16 years old on the date of their last assessment, then the SDQ score should be completed.  If the child has an SDQ score, then the SDQ score field should be completed and the reason for no SDQ score field should be left blank.  If no SDQ score is available, then the SDQ score field should be left blank and the reason for no SDQ score field should be completed in the range SDQ1 to SDQ5.
#### DfE suggested coding
If `<SDQ_SCORE>` has been provided then `<SDQ_REASON>` should not be provided

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

SLornieCYC:
Rule renamed 197**a** in the 2021 v1.2 validation rules: [Children looked after return 2021 to 2022: technical specifications](https://www.gov.uk/government/publications/children-looked-after-return-2021-to-2022-technical-specifications)

Issue #82, opened 2021-07-29
Error code 193 - Child not identified as having a substance misuse problem but at least one of the two additional items on whether an intervention were offered and received have been completed.


mwaddoups:
#### Error code
193
#### Description to display
Child not identified as having a substance misuse problem but at least one of the two additional items on whether an intervention were offered and received have been completed.
#### DfE Long-form description of error
If a child was not identified as having a substance misuse problem, then the two additional items on whether an intervention was offered or received should be left blank.  Data is not accepted in either the intervention received, or intervention offered fields once the substance misuse field is recorded as 0.
#### DfE suggested coding
if `<SUBSTANCE_MISUSE>` is  not provided then `<INTERVENTION_RECEIVED>` and `<INTERVENTION_OFFERED>` must also be missing

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

Issue #63, opened 2021-07-29
Error code 192 - Child has been identified as having a substance misuse problem but the additional item on whether an intervention was received has been left blank.


mwaddoups:
#### Error code
192
#### Description to display
Child has been identified as having a substance misuse problem but the additional item on whether an intervention was received has been left blank.
#### DfE Long-form description of error
If a child has been identified as having a substance misuse problem, then the additional item on whether an intervention was received should also be completed.  Only the item on whether an intervention was received must be completed once a child has been identified as having a substance misuse problem. The additional item on whether an intervention was offered is only required if the child was not offered an intervention for their substance misuse problem.
#### DfE suggested coding
Where `<SUBSTANCE_MISUSE>` is provided then `<INTERVENTION_RECEIVED>` must also be provided

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

PR #340, opened 2021-09-21
Added validation for error codes 153, 166, 174, 180, 181, 192, 193, 197

WMossDCC:
Validation rules added for the above error codes.

Closes issues #40 #46 #51 #57 #58 #63 (+ duplicate #81 ?) #82 #84


validator903/validators.py
```diff
@@ -1495,3 +1495,208 @@ def _validate(dfs):
...
+        if 'Reviews' not in dfs:
+            return {}
+        else:
+            review = dfs['Reviews']
+
+            mask = pd.to_datetime(review['REVIEW'], format='%d/%m/%Y', errors='coerce').notna()
+
+            na_location = review['REVIEW'].isna()
+
+            validation_error_mask = (~mask & ~na_location) | na_location
```
dezog, line 1550: The logic works out correct here but could be simplified: `validation_error_mask = ~mask` by itself should catch all cases, as with `errors='coerce'`, `to_datetime` results in `pd.NA` for both blank and invalid values, so there's no need to differentiate between them
```
>>> test_values = ['', 'NULL', np.nan, pd.NaT, pd.NA, 20210401, '2021-04-01', '99/04/2021', '01/04/2021']

>>> pd.to_datetime(test_values, format='%d/%m/%Y', errors='coerce').isna()

array([ True,  True,  True,  True,  True,  True,  True,  True,  False])
```
This change is optional, but I think it's worth making for the readability of the code

validator903/validators.py
```diff
@@ -1495,3 +1495,208 @@ def _validate(dfs):
...
+        affected_fields=['SDQ_SCORE'],
+    )
+
+    def _validate(dfs):
+        if 'OC2' not in dfs:
+            return {}
+        else:
+            oc2 = dfs['OC2']
+
+            mask = oc2['SDQ_SCORE'].astype(str).between('0','40')
```
dezog, line 1594: This is the right idea (and really nice and concise!), but this one's actually a bit trickier than it seems.

First, we need to compare these as integers, because the way string comparisons happen with `>` and `<` is basically in terms of which comes first alphabetically, so as long as the first character is  between 0 & 3, the rest could be anything:
```
 >>> pd.Series(['041', '09999', '3~abc~XYZ']).between('0', '40').values

array([ True,  True,  True])
```
I think the best way would be to use `pd.to_numeric(oc2['SDQ_SCORE'], errors='coerce')`, as doing `.astype(int)` directly would cause an error if it found a value it couldn't convert to an integer. 

The other thing is that since the rule requires a whole number, we need to flag up values like `30.5`, which means the `.between` method isn't enough on its own. You could either combine that with checking that the decimal part of the number is 0, or check against the allowed values directly, by doing `.isin([0, 1, 2, 3, (...) , 40])`, or equivalently, `.isin(range(41))`

tests/test_validators.py
```diff
@@ -905,3 +905,121 @@ def test_validate_502():
...
+
+    error_defn, error_func = validate_174()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'Header': [1, 5]}
+
+def test_validate_180():
+    fake_data = pd.DataFrame({
+        'SDQ_SCORE':    ['10', 41, '58', 72, '0', 40],
```
dezog, line 953: Adding few more exotic test cases would be good here, to catch the issues discussed in the comment on the validation check itself.

validator903/validators.py
```diff
@@ -1495,3 +1495,208 @@ def _validate(dfs):
...
+    )
+
+    def _validate(dfs):
+        if 'Header' not in dfs:
+            return {}
+        else:
+            header = dfs['Header']
+            code_list=['1']
+
+            mask = header['SEX'].astype(str).isin(code_list)
```
dezog, line 1571: optional change for conciseness - you could just simplify this to `mask = header['SEX'].astype(str) == '1'`

validator903/validators.py
```diff
@@ -1495,3 +1495,208 @@ def _validate(dfs):
...
+        affected_fields=['SUBSTANCE_MISUSE','INTERVENTION_RECEIVED'],
+    )
+
+    def _validate(dfs):
+        if 'OC2' not in dfs:
+            return {}
+        else:
+            oc2 = dfs['OC2']
+
+            misuse_na = oc2['SUBSTANCE_MISUSE'].isna()
```
dezog, line 1648: This appears correct according to the 'coding description', but the 'coding rule' is slightly contradictory and says this is supposed to checking if SUBSTANCE_MISUSE = 1 rather than if SUBSTANCE_MISUSE  is not blank. I'd tend toward the wording of the rule, "identified as having a substance misuse problem", what do you think?

image
image

validator903/validators.py
```diff
@@ -1495,3 +1495,208 @@ def _validate(dfs):
...
+        affected_fields=['SUBSTANCE_MISUSE','INTERVENTION_RECEIVED','INTERVENTION_OFFERED'],
+    )
+
+    def _validate(dfs):
+        if 'OC2' not in dfs:
+            return {}
+        else:
+            oc2 = dfs['OC2']
+
+            misuse_na = oc2['SUBSTANCE_MISUSE'].isna()
```
dezog, line 1671: The code looks perfect but there's the same ambiguity as Rule 192. Again the coding description states "if <SUBSTANCE_MISUSE> is not provided then (...)"

I am also confused by the part where it says "Data is not accepted in either the intervention received or intervention offered fields once the substance misuse field is recorded as 0. " Any idea if that means you don't have to check for that because it's not possible (so only check if SUBSTANCE_MISUSE is blank), rather than that it should be flagged up as an error?

validator903/validators.py
```diff
@@ -1495,3 +1495,208 @@ def _validate(dfs):
...
+    def _validate(dfs):
+        if 'OC2' not in dfs:
+            return {}
+        else:
+            oc2 = dfs['OC2']
+
+            score_na = oc2['SDQ_SCORE'].isna()
+            reason_na = oc2['SDQ_REASON'].isna()
+
+            validation_error_mask = ~score_na & ~reason_na
```
dezog, line 1697: This is spot on. 

Extremely minor style comment - it'd be slightly easier to follow to use `.notna()`, and then not have to negate the two masks

dezog (changes requested):
Great to see this come in. Excellent stuff and all very nicely laid out 👌 

Sorry for the delay in getting it back to you. 

I've put some comments throughout the code - the only definite issue's with rule 180, but it'd be great to get your point of view on the comments to rules 192 & 193, as the guidance document seems a bit ambiguous to me. 

The rest are just small style/readability points: optional changes.

Everything else looks perfect and is ready to go.

validator903/validators.py
WMossDCC: Hi, thank you for the feedback! I think I've made the required changes.. apart from this one. I'm honestly not sure. I think what I've put covers the error rule for blank entries, but like you say if it's recorded as 0 then does it void the other 2 fields?

SLornieCYC: Sorry, being nosy and looking at what everyone else is up to!

I think I'd expect 193 to work the same as discussed above for 192 and test whether <SUBSTANCE_MISUSE> = 0 rather than blank.

SUBSTANCE_MISUSE | INTERVENTION_RECEIVED | INTERVENTION_OFFERED | Outcome
----------------------- | ---------------------------- | ----------------------------- | ----------
0 | [null] | [null] | Correct
0 | [null] | _1 or 0_ | Error
0 | _1 or 0_ | [null] | Error
1 | [null] | [null] | Skip (Rule 192)
[null] | _1 or 0_ | [null] | Skip (Rule 182)
[null] | [null] | _1 or 0_ | Skip (Rule 182)
[null] | [null] | [null] | Skip (Rule 191 etc)

dezog: Not at all @SLornieCYC , the more eyes on issues like these the better! 

I can't see the changes @WMossDCC , have you pushed the commits to the remote copy of the branch? 

I also still feel slightly uncertain because of the weirdly clashing descriptions, but but I don't think we'll be able to get a clearer answer til the gov portal opens to compare... I agree with Stephen that it makes the most sense to only test the cases where 
- `<SUBSTANCE_MISUSE> = 1` for Rule 192 (_INTERVENTION_RECEIVED must not be null_)
- those where `<SUBSTANCE_MISUSE> = 0` for Rule 193 (_INTERVENTION_RECEIVED & INTERVENTION_OFFERED must both be null_)
-  with `<SUBSTANCE_MISUSE> = [null]` being covered by Rule 182, which is:

    > If any of <IMMUNISATIONS>; <TEETH_CHECK>; <HEALTH_ASSESSMENT> or  <SUBSTANCE MISUSE> have not been provided, then none of <CONVICTED>; <HEALTH_CHECK>; <INTERVENTION_RECEIVED> or <INTERVENTION_OFFERED> should be provided

Good detective work Stephen - by the way, how did you insert a table in a reply??

SLornieCYC: We can always contact DFE directly via their returns service desk and ask them to clarify the rule.  I've done that before outside the return period when I've spotted issues with the guidance.

Tables are written using markdown.  Not the quickest thing but okay for small tables: https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables

dezog: That's neat, didn't know you could do that. Thanks for the tip

Contacting DFE's a good idea, I suspect we'll find more rules that are a bit ambiguous at some point. Any idea where you can access that returns service desk now? The form linked in page 3 of the validation checks doc seems to be gone
image

In the meantime, shall we stick with the above?

SLornieCYC: I didn't even know that link was in the guide! 
Service desk is now here: https://form.education.gov.uk/service/Data-collections-service-request-form

(According to the latest return bulletin which arrived yesterday there's an update due to all of the return guidance "in October" so we'll have to see what that brings!)

dezog: Sweet, thanks. I'll request some clarification on this and any other similar issues I can find this week. 

That sounds like it'd be useful to be hearing about - how does one get access to these bulletins?

SLornieCYC: FYI that updated guidance has been published: [Children looked after return 2021 to 2022: technical specifications](https://www.gov.uk/government/publications/children-looked-after-return-2021-to-2022-technical-specifications)

Coincidentally one of the three changed rules is included in this PR! (197 renamed to 197a)

dezog: Ah, great, thanks for adding these in! Unfortunately the ones that were confusing us haven't been updated 😅

PR #343, opened 2021-09-28
Added validation rule for error 571.

### LINKED ISSUE #252: Error code 571 - The date that the child ceased to be missing or away from placement without authorisation is before the start or after the end of the collection year.
#### Error code
571
#### Description to display
The date that the child ceased to be missing or away from placement without authorisation is before the start or after the end of the collection year.
The date that the child ceased to be missing or away from placement without authorisation must be during the collection year.
#### DfE suggested coding
If provided `<MIS_END>` must NOT be `<= `<CURRENT_COLLECTION_YEAR>` -1 `<COLLECTION_END_DATE>` or >` `<CURRENT_COLLECTION_YEAR>` `<COLLECTION_END_DATE>`

SLornieCYC:
*Error does not trigger on 'missing' end dates (child may still be missing)
*Error does not trigger on invalid end dates (handled by separate validation rule)

Closes #252

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -1,6 +1,34 @@
 import pandas as pd
 from .types import ErrorDefinition
 
+def validate_571():
+    error = ErrorDefinition(
+        code = '571',
+        description = 'The date that the child ceased to be missing or away from placement without authorisation is before the start or after the end of the collection year.',
+        affected_fields=['MIS_END'],
+    )
+
+    def _validate(dfs):
+        if 'Missing' not in dfs:
+            return {}
+        else:
+            missing = dfs['Missing']
+            collection_start = pd.to_datetime(dfs['metadata']['collection_start'],format='%d/%m/%Y',errors='coerce')
+            collection_end = pd.to_datetime(dfs['metadata']['collection_start'],format='%d/%m/%Y',errors='coerce')
+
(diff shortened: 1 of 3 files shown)


validator903/validators.py
```diff
@@ -1,6 +1,34 @@
...
+
+    def _validate(dfs):
+        if 'Missing' not in dfs:
+            return {}
+        else:
+            missing = dfs['Missing']
+            collection_start = pd.to_datetime(dfs['metadata']['collection_start'],format='%d/%m/%Y',errors='coerce')
+            collection_end = pd.to_datetime(dfs['metadata']['collection_start'],format='%d/%m/%Y',errors='coerce')
+
+            missing['fMIS_END'] = pd.to_datetime(missing['MIS_END'], format='%d/%m/%Y', errors='coerce')
```
dezog, line 19: completely **optional**, but for simplicity & consistency with other rules I'd just use the original column name when doing type conversions like this, unless for some reason you also need access to the original text input

(You don't have to worry about overwriting the original values unless you need them within this rule, as each check gets provided with a copy of the uploaded dataframes, rather than direct access to them) 

I'll leave this branch intact incase you want to make the extra commit & start a new pull request for it, but no worries if not

dezog (approved):
Looks good, approved.

PR #348, opened 2021-10-05
Added error 574

### LINKED ISSUE #253: Error code 574 - A new missing/away from placement without authorisation period cannot start when the previous missing/away from placement without authorisation period is still open. Missing/away from placement without authorisation periods should also not overlap.
#### Error code
574
#### Description to display
A new missing/away from placement without authorisation period cannot start when the previous missing/away from placement without authorisation period is still open. Missing/away from placement without authorisation periods should also not overlap.
The previous missing/away from placement without authorisation period must have a valid end date before a new missing/away from placement without authorisation period can start. Whilst missing/away from placement without authorisation periods should not overlap, one missing/away from placement without authorisation period may end and another start on the same day.  Check that the previous missing/away from placement without authorisation period has a valid end date.  Check that missing/away from placement without authorisation periods do not overlap.
#### DfE suggested coding
If provided `<MIS_START>` from current Missing occurrence must be >`= previous Missing occurrence `<MIS_START>` and previous Missing occurrence `<MIS_END>` cannot be Null

PhilDeakin:
Added 574.

Note in the coding description it reads:
If provided <MIS_START> from current Missing occurrence must be >= previous Missing occurrence <MIS_START> and
previous Missing occurrence <MIS_END> cannot be Null

It should be <MIS_START> must be >= previous <MIS_END> and <MIS_END> can't be null , otherwise it doesn't make sense.

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -1101,3 +1101,24 @@ def test_validate_1011():
     result = error_func(fake_dfs)
 
     assert result == {'OC3': [0,2,4]}
+
+def test_validate_574():
+    fake_mis = pd.DataFrame([
+    {'CHILD': '111', 'MIS_START' : '01/06/2020', 'MIS_END' : '05/06/2020' }, #0  
+    {'CHILD': '111', 'MIS_START' : '04/06/2020', 'MIS_END' : pd.NA },        #1 Fails overlaps
+    {'CHILD': '222', 'MIS_START' : '03/06/2020', 'MIS_END' : '04/06/2020' }, #2 
+    {'CHILD': '222', 'MIS_START' : '04/06/2020', 'MIS_END' : pd.NA },        #3 
+    {'CHILD': '222', 'MIS_START' : '07/06/2020', 'MIS_END' : '09/06/2020' }, #4 Fails, previous end is null 
+    {'CHILD': '333', 'MIS_START' : '02/06/2020', 'MIS_END' : '04/06/2020' }, #5
+    {'CHILD': '333', 'MIS_START' : '03/06/2020', 'MIS_END' : '09/06/2020' }, #6 Fails overlaps
+    {'CHILD': '555', 'MIS_START' : pd.NA,        'MIS_END' : '05/06/2020' }, #7  
+    {'CHILD': '555', 'MIS_START' : pd.NA,        'MIS_END' : '05/06/2020' }, #8  
+    ])
+
+    fake_dfs = {'Missing': fake_mis}
+
+    error_defn, error_func = validate_574()
+
+    result = error_func(fake_dfs)
+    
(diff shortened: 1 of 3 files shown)


validator903/validators.py
```diff
@@ -1776,4 +1776,45 @@ def _validate(dfs):
...
+        else:
+  
+            mis = dfs['Missing']  
+            mis['MIS_START'] = pd.to_datetime(mis['MIS_START'], format='%d/%m/%Y', errors='coerce')
+            mis['MIS_END'] = pd.to_datetime(mis['MIS_END'], format='%d/%m/%Y', errors='coerce')
+            
+            mis.sort_values(['CHILD','MIS_START'],inplace=True)
+
+            mis.reset_index(inplace=True)
+            mis.reset_index(inplace=True)  #Twice on purpose
```
dezog, line 1800: Am I missing something here, or couldn't you do this once and use the 'index' column in place of 'level_0' just the same?

tests/test_validators.py
```diff
@@ -1100,4 +1100,25 @@ def test_validate_1011():
...
\ No newline at end of file
+    assert result == {'OC3': [0,2,4]}
+
+def test_validate_574():
+    fake_mis = pd.DataFrame([
+    {'CHILD': '111', 'MIS_START' : '01/06/2020', 'MIS_END' : '05/06/2020' }, #0  
+    {'CHILD': '111', 'MIS_START' : '04/06/2020', 'MIS_END' : pd.NA },        #1 Fails overlaps
+    {'CHILD': '222', 'MIS_START' : '03/06/2020', 'MIS_END' : '04/06/2020' }, #2 
+    {'CHILD': '222', 'MIS_START' : '04/06/2020', 'MIS_END' : pd.NA },        #3 
+    {'CHILD': '222', 'MIS_START' : '07/06/2020', 'MIS_END' : '09/06/2020' }, #4 Fails, previous end is null
```
dezog, line 1111: love these little comments, super helpful 🥇

dezog (approved):
Hey Phil, very nice rule here, approved as is.

validator903/validators.py
PhilDeakin: I thought the "left_on" , "right_on" parameters of the merge had to be column name strings, but seemingly not, it'll have   left_on= mis.index. 
Doesn't really matter now, cheers.

dezog: Interesting - so did I! --but I just meant to say I reckoned passing `left_on='index', right_on='LAG_INDEX` to the merge would have the same effect.

But I suppose it might not, if the index isn't just `[0, 1, 2, 3, (...)]`:
image

This reminds me though, I'd been meaning to look into making sure that it is at the point we load the files in, as some of the rules (not this one though!) rely on that assumption

PR #353, opened 2021-10-15
Added error 436

### LINKED ISSUE #196: Error code 436 - Reason for new episode is that both child's placement and legal status have changed, but this is not reflected in the episode data.
#### Error code
436
#### Description to display
Reason for new episode is that both child's placement and legal status have changed, but this is not reflected in the episode data.
Where the reason for new episode is coded B or U the legal status of the current episode must be different from the previous episode, and at least one of placement postcode, placement URN, placement type, or placement provider must also be different.  Check the legal status and placement codes in current and previous episodes.
#### DfE suggested coding
Where `<RNE>` = 'B' or 'U' then `<LS>` must not = previous episode `<LS>` and at least one of `<PL>`, `<PL_POSTCODE>`, `<URN>` or `<PLACE_PROVIDER>` must be different

PhilDeakin:
Added error 436


dezog (approved):
Slightly fiddly one this, well done - approved

PR #361, opened 2021-10-21
Added 431

### LINKED ISSUE #191: Error code 431 - The reason for new episode is started to be looked after, but the previous episode ended on the same day.
#### Error code
431
#### Description to display
The reason for new episode is started to be looked after, but the previous episode ended on the same day.
If reason for new episode is coded S the episode cannot start on the same day as the previous episode ended. Check that the last episode ended, and the current episode started on the same day. If so, use reason new episode code P, L, T, U or B as appropriate.
#### DfE suggested coding
If `<RNE>` = 'S' then `<DECOM>` must be greater than `<DEC>` of previous episode

PhilDeakin:
Added 431


validator903/validators.py
```diff
@@ -1938,3 +1938,32 @@ def _validate(dfs):
 
     return error, _validate
 
+def validate_431():
+    error = ErrorDefinition(
+        code='431',
+        description='The reason for new episode is started to be looked after, but the previous episode ended on the same day.',
+        affected_fields=['RNE','DECOM','DEC'],
```
dezog: I think DEC should be dropped from `affected_fields`, as it'll highlight the DEC of the wrong row in the frontend (it's the DECOM of the error row and the DEC of the previous row that are in conflict)

It would be nice to track separate error indices for each column for rules like this, but that'd be a big job considering the timeframes

dezog (changes requested):
I've highlighted a change to the affected_fields, let me know if you agree!

dezog (approved):

PR #365, opened 2021-10-27
Added validator for error rule 204.

### LINKED ISSUE #109: Error code 204 - Ethnic origin code disagrees with the ethnic origin already recorded for this child.
#### Error code
204
#### Description to display
Ethnic origin code disagrees with the ethnic origin already recorded for this child.
DfE holds an existing record for this child, but the ethnic origin details do not match.  If you are certain that these details are correct and previous details are wrong, please amend through the master change function or contact DfE about correcting earlier years.
#### DfE suggested coding
`<ETHNIC>` from current year's record must = `<ETHNIC>` from previous year's record

SLornieCYC:
Closes #109


validator903/validators.py
```diff
@@ -1,6 +1,33 @@
...
+        if 'Header' not in dfs or 'Header_last' not in dfs:
+            return {}
+        else:
+            header = dfs['Header']
+            header_last = dfs['Header_last']
+
+            header_merged = header.reset_index().merge(header_last, how='left', on=['CHILD'], suffixes=('', '_last'), indicator=True).set_index('index')
+
+            in_both_years = header_merged['_merge'] == 'both'
+            ethnic_is_different = header_merged['ETHNIC'].astype(str).str.upper() != header_merged['ETHNIC_last'].astype(str).str.upper()
```
dezog, line 21: Just to give you a heads up, we've decided to just convert all text fields (except UPN, as it mentions case in the rule) to upper case on reading in the files, so this shouldn't be necessary going forward 🙂 

(unless you can think of any case sensitive rules other than rule 168 about UPN's?)

validator903/validators.py
```diff
@@ -1,6 +1,33 @@
...
+    )
+
+    def _validate(dfs):
+        if 'Header' not in dfs or 'Header_last' not in dfs:
+            return {}
+        else:
+            header = dfs['Header']
+            header_last = dfs['Header_last']
+
+            header_merged = header.reset_index().merge(header_last, how='left', on=['CHILD'], suffixes=('', '_last'), indicator=True).set_index('index')
```
dezog, line 18: Free style tip - lines like these where you're applying multiple methods in a row, you can typeset like so for readability:
```
header_merged = (
    header
        .reset_index()
        .merge(header_last, how='left', on=['CHILD'], suffixes=('', '_last'), indicator=True)
        .set_index('index')
)
```
Just an idea; up to you what reads best in any given case!

dezog (approved):
No complaints, approved!

PR #368, opened 2021-10-27
Added 380 & 381

### LINKED ISSUE #167: Error code 380 - A period of care cannot start with a temporary placement.
#### Error code
380
#### Description to display
A period of care cannot start with a temporary placement.
Where the placement code is in the range T0 to T4 the reason for new episode code must be P or B.  A temporary placement, by definition, is one where there is clear expectation from the outset that a child will return to his/her normal placement and that this expectation is duly fulfilled.
#### DfE suggested coding
Where `<PL>` = 'T0' or 'T1' or 'T2' or 'T3' or 'T4' then `<RNE>` must be 'P' or 'B'

### LINKED ISSUE #168: Error code 381 - A period of care cannot end with a temporary placement.
#### Error code
381
#### Description to display
A period of care cannot end with a temporary placement.
Where the placement code is in the range T0 to T4 the reason episode ceased must be X1.  A temporary placement is, by definition, one where there is clear expectation from the outset that a child will return to his/her normal placement, and that this expectation is duly fulfilled.
#### DfE suggested coding
Where `<PL>` = 'T0' or 'T1' or 'T2' or 'T3' or 'T4' then `<REC>` must be 'X1'

PhilDeakin:
As above


tests/test_validators.py
```diff
@@ -1339,3 +1339,30 @@ def test_validate_562():
...
+        'PLACE': ['T3', 'T0', 'U2',  'T2', 'T1', pd.NA], 
+    })
+
+    fake_dfs = {'Episodes': fake_data}
+    
+    error_defn, error_func = validate_381()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'Episodes': [1, 3]}
```
SLornieCYC: I don't think case 3 should fail this test (381)?  If there's no `<REC>` that means the placement (and care period) is still open so I think at the moment this will treat any active Tx placement as being in-error. (also PR1 isn't a valid `<REC>` but that's not strictly relevant for this test)

validator903/validators.py
```diff
@@ -2077,4 +2077,37 @@ def _validate(dfs):
...
+        description='A period of care cannot start with a temporary placement.',
+        affected_fields=['PLACE','RNE'],
+    )
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        else:
+            epi = dfs['Episodes']  
+            error_mask = (epi['PLACE'].isin(['T0','T1','T2','T3','T4'])) & (~epi['RNE'].isin(['P','B']))
```
dezog: Extremely minor style note: you could drop the brackets around the two `.isin`'s on this line, as they're only needed when, for example:
- you are using an operator like `==` or `<` and need to stop python trying to make the wrong comparison:
```
(epi['PLACE'] == 'T0') & (epi['RNE'] == 'P')
```
- you need to combine them into a more complex piece of logic, such as:
```
(epi['PLACE'].isin(list_A) & epi['REC'].isin(list_B)) | epi['RNE'].isna()
```

dezog (approved):
Looks good! 👍

PR #371, opened 2021-10-29
Added 354

### LINKED ISSUE #145: Error code 354 - The episode starts after the end of the current collection year.
#### Error code
354
#### Description to display
The episode starts after the end of the current collection year.
All episodes on this record must begin on or before 31 March of the collection year to which this record refers â€“ 1 April to 31 March of the current collection year.  Check the date episode commenced. If it is correct, should this episode be reported on next year's record instead?
#### DfE suggested coding
`<DECOM>` must be `<= `<COLLECTION_END_DATE>`

DMC63:
Added 354


tests/test_validators.py
```diff
@@ -1339,3 +1339,16 @@ def test_validate_562():
 
     assert result == {'Episodes': [2, 5]}
 
+def test_validate_354():
+    fake_data = pd.DataFrame({'DECOM': ['01/01/2021', '19/02/2010', '38/04/2019', '01/01/2022'],})
```
dezog: I would add a couple of extra test cases here, just for peace of mind: 
- `pd.NA`, to ensure missing values don't flag up this error
- an incorrectly formatted date, likewise
- the collection date itself and the day immediately after, to make sure we're handling that correctly

dezog (changes requested):
Nice one David, the rule itself looks perfect. I would just add some test cases for completeness, as described in the comment below. 

They won't make any practical difference here, but it's good practice for more complicated rules where it's really easy for edge cases to slip through the net

tests/test_validators.py
dezog: Going to add these and merge, as everything else is great - good to bear these cases in mind for these date-comparison rules though

dezog (approved):

PR #372, opened 2021-10-29
Added 504

### LINKED ISSUE #208: Error code 504 - The category of need code differs from that reported at start of current period of being looked after.
#### Error code
504
#### Description to display
The category of need code differs from that reported at start of current period of being looked after.
Where the current episode directly follows a previous episode the category of need code must be the same in both episodes.  The category of need code recorded is that at the start of the current period of being looked after. So long as the child continues to be looked after without a break in care the same category of need code must be recorded on the data collection return.
#### DfE suggested coding
Where previous episode `<REC>` = 'X1' then `<CIN>` must = previous episode `<CIN>`

PhilDeakin:
The test_integration part of pytest is failing on this and I can't work out why.

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -2078,3 +2078,32 @@ def validate_562():
 
     return error, _validate
   
+def validate_504():
+    error = ErrorDefinition(
+        code='504',
+        description='The category of need code differs from that reported at start of current period of being looked after',
+        affected_fields=['CIN'],
+    )
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        else:
+            epi = dfs['Episodes']
+            epi['DECOM'] = pd.to_datetime(epi['DECOM'], format='%d/%m/%Y', errors='coerce')
+            epi['DEC'] = pd.to_datetime(epi['DEC'], format='%d/%m/%Y', errors='coerce')
+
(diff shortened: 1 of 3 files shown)


validator903/validators.py
```diff
@@ -2077,4 +2077,33 @@ def _validate(dfs):
...
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        else:
+            epi = dfs['Episodes']
+            epi['DECOM'] = pd.to_datetime(epi['DECOM'], format='%d/%m/%Y', errors='coerce')
+            epi['DEC'] = pd.to_datetime(epi['DEC'], format='%d/%m/%Y', errors='coerce')
+
+            epi.sort_values(['CHILD','DECOM'],inplace=True)
+            epi.reset_index(inplace=True)
```
dezog, line 2097: I'm still pretty sure you could get away with just resetting the index once and then using the 'index' column instead of 'level_0'

the result should be the same, as the site doesnt read in an index column from the .csv's, meaning the index should in any case always start off as a simple range: (0, 1, 2, 3, ...)

dezog (approved):
as discussed, the integration tests were failing due to a bug in the tests themselves. I've now fixed that and this is looking all good, cheers Phil!

PR #376, opened 2021-11-03
Added validator for rule 387.

### LINKED ISSUE #174: Error code 387 - Reason episode ceased is child moved into independent living arrangement, but the child is aged under 14.
#### Error code
387
#### Description to display
Reason episode ceased is child moved into independent living arrangement, but the child is aged under 14.
If the reason episode ceased is coded E5 or E6 the age at date episode ceased should be 14th birthday or over.  Check reason episode ceased code and date of birth.
#### DfE suggested coding
If `<REC>` = 'E5' or 'E6' then `<DOB>` must be >`=14 years prior to `<DEC>`

SLornieCYC:
Closes #174


validator903/validators.py
```diff
@@ -1,6 +1,40 @@
...
+
+            header['DOB'] = pd.to_datetime(header['DOB'], format='%d/%m/%Y', errors='coerce')
+            episodes['DEC'] = pd.to_datetime(episodes['DEC'], format='%d/%m/%Y', errors='coerce')
+            header['DOB14'] = header['DOB'] + pd.DateOffset(years=14)
+
+            episodes_merged = episodes.reset_index().merge(header, how='left', on=['CHILD'], suffixes=('', '_header'), indicator=True).set_index('index')
+
+            ceased_indep = episodes_merged['REC'].str.upper().astype(str).isin(['E5','E6'])
+            ceased_over_14 = episodes_merged['DOB14'] <= episodes_merged['DEC']
+            dec_present = episodes_merged['DEC'].notna()
```
dezog, line 28: added this small tweak & a couple test cases to make sure missing/invalid DEC's don't trigger this error

dezog (approved):
Good work, merging now.

Made a small change noted below

validator903/validators.py
SLornieCYC: Thanks! I was relying on `DOB14 <= DEC` returning false if DEC is null (following the to_datetime coerce).

dezog: Ah - I figured the resulting mask gets flipped with `~`, so they'd end up counting as true. One to watch out for!

PR #381, opened 2021-11-05
Added 552

### LINKED ISSUE #235: Error code 552 - Date of decision to place a child for adoption should be on or prior to the date that the child was placed for adoption.
#### Error code
552
#### Description to display
Date of decision to place a child for adoption should be on or prior to the date that the child was placed for adoption.
A child cannot be placed for adoption until the local authority has made the formal decision that the child should be placed for adoption.
#### DfE suggested coding
If `<PL>` = 'A3' or 'A4' or 'A5' or 'A6' then `<DATE_PLACED>` must be `<= `<DECOM>`

tab1tha:
Added rule 552

Closes #235


tests/test_validators.py
```diff
@@ -1,6 +1,24 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_552():
+  fake_placed_adoption = pd.DataFrame({
+    'DATE_PLACED':['08/03/2020','22/06/2020','13/10/2022','24/10/2021'],
```
dezog: What happened to the tests we had added including missing values? 

Currently, if the user loads a file with missing values in DATE_PLACED or DECOM this function will break, so we need to put it back to how it was in https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/pull/381/commits/4dcbfa8cfd03a369ea62ff48c7f0d4365a06fa83, and add the changes to validators.py (see other comment)

validator903/validators.py
```diff
@@ -1,6 +1,54 @@
...
+      episodes = dfs['Episodes']
+
+      # keep index values so that they stay the same when needed later on for error locations
+      placed_adoption.reset_index(inplace = True)
+      episodes.reset_index(inplace = True)
+
+      adoption_eps = episodes[episodes['PLACE'].isin(['A3', 'A4', 'A5', 'A6'])]
+      # find most recent adoption decision
+      placed_adoption['DATE_PLACED'] = pd.to_datetime(placed_adoption['DATE_PLACED'], format='%d/%m/%Y', errors='coerce')
+      placed_adoption_inds = placed_adoption.groupby("CHILD")["DATE_PLACED"].idxmax(skipna=True)
```
dezog: We need to remove the rows of `placed_adoption` where `DATE_PLACED.isna()` before this line, to stop idxmax running into problems

validator903/validators.py
```diff
@@ -1,6 +1,54 @@
...
+      last_decision = placed_adoption.loc[placed_adoption_inds]
+
+      # first time child started adoption
+      adoption_eps["DECOM"] = pd.to_datetime(adoption_eps['DECOM'], format='%d/%m/%Y', errors='coerce')
+      adoption_eps_inds = adoption_eps.groupby('CHILD')['DECOM'].idxmin(skipna=True)
```
dezog: Like above, we should also remove the rows of `adoption_eps` where DECOM's a NaN

poetry.lock
```diff
@@ -46,25 +46,30 @@ python-versions = ">=3.7"
```
dezog, line 46: any idea what happened here? did you update the package versions in the poetry file somehow?

dezog (changes requested):
Hey Tambe, 

this is really close to done now - still a couple of changes to make; see the comments below.

This was a really difficult and confusing rule to start off with, so well done for getting it into shape

poetry.lock
tab1tha: I have no idea yet.

validator903/validators.py
tab1tha: Done

validator903/validators.py
tab1tha: Done

tests/test_validators.py
tab1tha: Done

validator903/validators.py
dezog: Thanks for putting these in, it's working fine for our test cases, but I've noticed a bug: `placed_adoption = dfs['PlacedAdoption'].dropna()` will drop the rows where *any* column contains a NaN. 

Adding `subset='DATE_PLACED` solves this, but since using `to_datetime` with `errors='coerce'` will replace invalid values with NaN's, we should use dropna after that happens:
```
placed_adoption = dfs['PlacedAdoption']
# ...
placed_adoption['DATE_PLACED'] = pd.to_datetime(placed_adoption['DATE_PLACED'], format='%d/%m/%Y', errors='coerce') 
placed_adoption = placed_adoption.dropna(subset='DATE_PLACED')
# ...
```
and likewise with adding `subset='DECOM'`  for episodes

dezog (changes requested):
Nice work, just one more tiny fix to make!

validator903/validators.py
tab1tha: using subset did not produce the required results. This worked instead: https://stackoverflow.com/questions/13413590/how-to-drop-rows-of-pandas-dataframe-whose-value-in-a-certain-column-is-nan

dezog: Completely my bad, sorry about that - the subset needs to be a list, e.g. `placed_adoption.dropna(subset=['DATE_PLACED'])`

tab1tha: Would you like me to change it back to this instead?

dezog (approved):

PR #382, opened 2021-11-08
Added rule for error 448 (issue 208)

Steve-Far:
Coded rule and test, please check


dezog (approved):
Perfect first rule, nice one Steve!

PR #388, opened 2021-11-10
210 validate 514

OgileOze:
Update to 514 validation rule


tests/test_validators.py
```diff
@@ -1,6 +1,20 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_210():
+    fake_data = pd.DataFrame({
+        'LS_ADOPTR':  ['M1', 'F1', 'M2', pd.NA],
+    })
```
dezog: So we need to include the two columns we're interested in in `fake_data` - `'LS_ADOPTR'`, and `'SEX_ADOPTR'`, and the rows where LS_ADOPTR is 'L0', but SEX_ADOPTR is neither 'M1' or 'F1' should be the errors. 

For example, in the below rows 2 & 3 should be errors as the LS is 'L0', but the SEX is missing or wrong. 4 & 5 the SEX code is wrong but since the LS is not 'L0', it shouldn't cause this error

```
    fake_data = pd.DataFrame({
        'SEX_ADOPTR': ['M1', 'F1', 'XX', pd.NA, 'XX', 'XX'],
        'LS_ADOPTR': ['L0', 'L0', 'L0', 'L0', 'LX', pd.NA],
                       #0    #1    #2    #3    #4     #5
    })
```

tests/test_validators.py
```diff
@@ -1,6 +1,20 @@
...
+  
+    fake_dfs = {'AD1': fake_data}
+
+    error_defn, error_func = validate_210()
+
+    result = error_func(fake_dfs)
+
+    assert result == {'AD1': [2, 3]}
```
dezog: This is all good :+1:

validator903/validators.py
```diff
@@ -1,6 +1,32 @@
 import pandas as pd
 from .types import ErrorDefinition
 
+def validate_210():
+    error = ErrorDefinition(
+      code='210',
+      description= 'Data entry on the legal status of adopters shows a single adopter but data entry for the numbers of adopters shows it as a couple.',
+      affected_fields=['LS_ADOPTR'],
```
dezog: Don't forget to include SEX_ADOPTR here

validator903/validators.py
```diff
@@ -1,6 +1,32 @@
...
+
+    def _validate(dfs):
+
+        if 'AD1' not in dfs:
+          return {}
+        else:
+            AD1 = dfs ['AD1']
+            code_list = ['M1', 'F1']
+
+            mask = AD1['LS_ADOPTR'].isin(code_list)
```
dezog: Easy to get these mixed up - the codes M1 & F1 are for SEX_ADOPTR, so that should change on line 19

validator903/validators.py
```diff
@@ -1,6 +1,32 @@
...
+
+            validation_error_mask = ~mask
```
dezog: After fixing line 19, `validation_error_mask` currently contains `True` for all rows of AD1 where the SEX_ADOPTR code is M1 or F1

We need to combine this with another mask to get `True` for only the rows where SEX_ADOPTR is M1 or F1, and the LS_ADOPTR column is also L0, like so:
(note the use of brackets)

```
validation_error_mask = ~mask & (AD1['LS_ADOPTR'] == 'L0')
```

I usually prefer to put the masks in descriptively named variables as I find it less confusing, especially when things get more complex. So you could alternatively do like this for the same result:
```
LS_is_L0 = AD1['LS_ADOPTR'] == 'L0'
SEX_not_M1_F1 = ~AD1['SEX_ADOPTR'].isin(code_list)
validation_error_mask = LS_is_L0  & SEX_not_M1_F1

Comment (dezog) [reply validator903/validators.py]: Both ways are totally fine here though - up to you!

File: validator903/validators.py [hunk 6]
```diff
@@ -1,6 +1,32 @@
...
+
+            validation_error_locations = AD1.index[validation_error_mask]
+
+            return {'AD1': validation_error_locations.to_lists()}
```
dezog: tiny typo causing an error here - this should be `.to_list()` - no 's' at the end

dezog (changes requested):
Hey Oze, nice work here, good first attempt but a few changes to make. 

Tha main one is we need to use two columns of the AD1 table LS_ADOPTR and SEX_ADOPTR, and record as errors the rows where the LS_ADOPTR is L0, but SEX_ADOPTR is neither M1 or F1

image

I've put more detailed comments in the code below. Let me know if anything's not clear 🙂

dezog (approved):

dezog:
Great work guys!

PR #389, opened 2021-11-10
Added rule for issues 199 (Error code 441)

### LINKED ISSUE #199: Error code 441 - Date of birth and review date are inconsistent with participation method.
#### Error code
441
#### Description to display
Date of birth and review date are inconsistent with participation method.
Participation method indicates child was 4 years old or over at the time of the review, but the date of birth and review date indicates the child was under 4 years old.
#### DfE suggested coding
If `<REVIEW_CODE>` = 'PN1', 'PN2', 'PN3', 'PN4', 'PN5', 'PN6' or 'PN7' then `<DOB>` should be 4 years or more prior to `<REVIEW>`

Steve-Far:
Please review..


validator903/validators.py
```diff
@@ -1,6 +1,31 @@
...
+
+    def _validate(dfs):
+        if 'Reviews' not in dfs:
+          return {}
+        else:
+            reviews = dfs['Reviews']
+            reviews['DOB'] = pd.to_datetime(reviews['DOB'],format='%d/%m/%Y',errors='coerce')
+            reviews['REVIEW'] = pd.to_datetime(reviews['REVIEW'],format='%d/%m/%Y',errors='coerce')
+            
+            mask = reviews['REVIEW_CODE'].ne('PN0') & (reviews['REVIEW'] < reviews['DOB'] + pd.offsets.DateOffset(years=4))
```
dezog: One more small finnicky change - I'd swap:
`reviews['REVIEW_CODE'].ne('PN0')` 
...to... 
`reviews['REVIEW_CODE'].isin(['PN1’, ‘PN2’, ‘PN3’, ‘PN4’, ‘PN5’, ‘PN6’, ‘PN7’])`

This is just so this specific rule doesn't get triggered by incorrect or missing values of REVIEW_CODE, alongside error code 167 which would be more relevant in that case.

validator903/validators.py
```diff
@@ -1,6 +1,31 @@
...
+        affected_fields=['DOB','REVIEW','REVIEW_CODE'],
+    )
+
+    def _validate(dfs):
+        if 'Reviews' not in dfs:
+          return {}
+        else:
+            reviews = dfs['Reviews']
+            reviews['DOB'] = pd.to_datetime(reviews['DOB'],format='%d/%m/%Y',errors='coerce')
+            reviews['REVIEW'] = pd.to_datetime(reviews['REVIEW'],format='%d/%m/%Y',errors='coerce')
```
dezog, line 17: After setting DOB & REVIEW to datetimes, it would be good to drop the rows where either of these is a Null value, since using `to_datetime` with `errors='coerce'` will change invalid values to Nulls, and those should trigger a different error. You could do this by adding the following line:
```
reviews = reviews.dropna(subset=['REVIEW', 'DOB'])
```

dezog: In this case I don't think it'll make any difference, (I see you included a pd.NA test case, nice one!) so **this change is optional** 

However, if you know those rows aren't covered by an error it's often best to remove them at some point as they can slip through - for example, if instead of checking `REVIEW < DOB`, you checked `REVIEW >= DOB`, then negated the resulting mask with `~`, those rows would be unexpectedly marked as True!

dezog (commented):
Nice work Steve, just one or two small changes to make 🙂

dezog (changes requested):
a couple changes requested - see other comments

validator903/validators.py
Steve-Far: Hi Desi, when I change this it throws up a syntax error, any ideas on how to resolve?

 mask = reviews['REVIEW_CODE'].isin(['PN1’, ‘PN2’, ‘PN3’, ‘PN4’, ‘PN5’, ‘PN6’, ‘PN7’]) & (reviews['REVIEW'] < reviews['DOB'] + pd.offsets.DateOffset(years=4))

Steve-Far: Ignore that, it was the quotation marks around PN2 etc they were different characters.. sorted now

dezog: Ah, good thing to be aware of - IIRC the default text editors in windows & mac both like to do 'clever' typesetting stuff sometimes which can totally break code while looking fine to the untrained eye. (fun times.... 🙂)

dezog (approved):
Sorted! merging now

PR #391, opened 2021-11-11
Added rule for issue 209 (error code 511)

Steve-Far:
Please review


validator903/validators.py
```diff
@@ -1,6 +1,29 @@
...
+    )
+
+    def _validate(dfs):
+        if 'AD1' not in dfs:
+          return{}
+
+        else:
+            AD1 = dfs['AD1']
+
+            mask = AD1['NB_ADOPTR'].eq('2') & AD1['SEX_ADOPTR'].isin(['M1','F1'])
```
dezog: Changing this to match the wording "(...) then the code should only be MM, FF or MF"

dezog (approved):

validator903/validators.py
dezog: Otherwise this is great, going to merge it now :)

dezog: Oh, also adding `.astype(str)` in here, so it includes when the `NB_ADOPTR` is recorded as an integer (`2`) rather than a string (`'2'`):

```
AD1['NB_ADOPTR'].astype(str).eq('2')
```
Adding a couple of test cases to this effect as well

dezog:
Made a couple little tweaks, see comments above for explanations. You can also select the 'Files changed' tab above, then select my commit to see the changes in the code

image

PR #392, opened 2021-11-11
Rule added issue 211 (error 516)

Steve-Far:
Please Review


validator903/validators.py
```diff
@@ -1,6 +1,32 @@
...
+        affected_fields=['REC','PLACE'],
+    )
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+          return{}
+
+        else:
+            episodes = dfs['Episodes']
+            bad_place_codes = ['A3','A4','A5','A6','H5','K1','K2', 'P1', 'P2', 'P3', 'R1', 'R2', 'R3', 'R5', 'S1', 'T0', 'T1', 'T2', 'T3', 'T4', 'Z1']
```
dezog: Here I think it makes more sense to throw an error for any value except the right ones (U1 to U6)

image

dezog (approved):
Great work this week Steve, this looks good, just making one small change to reflect how its worded in the spec.

Issue #207, opened 2021-07-29
Error code 503 (amended) - The reason for new episode, legal status, category of need code, placement type, placement provider, placement Ofsted URN, distance or placement location in first episode do not match open episode at end of last year.


mwaddoups:
#### Error code
503 (amended)
#### Description to display
The reason for new episode, legal status, category of need code, placement type, placement provider, placement Ofsted URN, distance or placement location in first episode do not match open episode at end of last year.
#### DfE Long-form description of error
Reason for new episode, legal status, category of need code, placement type, placement provider, placement Ofsted URN, distance, local authority placement or placement location in first episode must be consistent with equivalent details in the last episode of a previous year's record which was open at the end of the year.  Check details for reason for new episode, legal status and placement are correct. If they are correct, are the details in last year's final episode correct? Were there one or more intervening episodes which have been left out?  It may be necessary to enter additional episode details into the system.  This applies to the 2016 to 2017 data collection onwards.
#### DfE suggested coding
Where previous collection year's final episode `<DEC>` not provided then first episode of this year's collection `<RNE>`, `<LS>`, `<CIN>`, `<PLACE_PROVIDER>`, `<OFSTED_URN>` and `<PL_DISTANCE>` must be the same as the previous episode  and  `<PL>` must = previous episode `<PL>`  and  `<PL_LA>` and `<PL_LOCATION>` must be the same as the previous episode

For further info, refer to the DfE validation document [here](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974999/Children_looked_after_data_collection_2020_to_2021_validation_checks_-_version_1.3.pdf)

PR #396, opened 2021-11-12
Added validator for rule 453. Also tweaked rule 452 and updated tests.

### LINKED ISSUE #205: Error code 453 - Contradiction between placement distance in the last episode of the previous year and in the first episode of the current year.
#### Error code
453
#### Description to display
Contradiction between placement distance in the last episode of the previous year and in the first episode of the current year.
Where a placement has not changed as at 31 March in the previous year, the placement distance should match the first episode of the current year.  Check placement code, placement location and local authority of placement are correct.
#### DfE suggested coding
Where `<RNE>` of first episode in current year = `<RNE>` of last episode of previous year and `<DEC>` is Null then `<PL_DISTANCE>` of both episodes should be the same

SLornieCYC:
Closes #205


SLornieCYC:
~~Still need to add 503 part J (#207) which also fits closely with the above, so makes sense to do at same time.~~ Now added.

dezog (approved):
Hey Stephen, I've made a couple of changes to this one:
- Separated out the test functions, while still letting them use the same test data. Not a big thing, having a separate function for each validation check just makes it a little easier for me to keep track and make sure everything's covered by a test
- Removed rows where DECOM.isna() before as in another rule we ran into a problem where this could cause idxmin & idxmax to raise an exception where there was a group with only NaN's in the sort column

PR #397, opened 2021-11-12
Added rule for 551 - 

### LINKED ISSUE #234: Error code 551 - Child has been placed for adoption but there is no date of the decision that the child should be placed for adoption.
#### Error code
551
#### Description to display
Child has been placed for adoption but there is no date of the decision that the child should be placed for adoption.
If a child has a placement code of A3 to A6 then the date of the decision that the child should be placed for adoption must be completed.  If the child has been placed for adoption, then complete the date of the decision that the child should be placed for adoption.  The date completed should be on or before the date the local authority formally decided that the child should be placed for adoption.  This rule applies to all records from 1 April 2008.
#### DfE suggested coding
If `<PL>` = 'A3' or 'A4' or 'A5' or 'A6' then `<DATE_PLACED>` must be provided

Steve-Far:
Hi Desi  - Am struggling with this one and need some help with the code!


validator903/validators.py
```diff
@@ -1,6 +1,40 @@
...
+
+            episodes = episodes.reset_index()
+
+            place_codes = ['A3','A4','A5','A6']
+
+            placeEpisodes = episodes[episodes['PLACE'].isin (place_codes)]
+
+            merged = placeEpisodes.reset_index().merge(placedAdoptions, how='left', on='CHILD', ).set_index('index')
+
+            episodes_with_errors = merged[merged['DATE_PLACED'] == 'NaN']
```
dezog: to check for a NaN, use the pandas method `.isna()` (a.k.a. `.isnull()` - same thing) - currently this checks if it contains the text `'NaN'`, rather than actual 'null values' as far as Python's concerned:
```
episodes_with_errors = merged[merged['DATE_PLACED'].isna()]
```
I like the way you've done this as it will flag both rows where `placedAdoptions.DATE_PLACED.isna()` is true, and those where the child isn't recorded in the placedAdoptions table, as after the merge they'll also have missing values in the DATE_PLACED column.

validator903/validators.py
```diff
@@ -1,6 +1,40 @@
...
+    )
+
+    def _validate(dfs):
+        if 'Episodes' not in dfs or 'PlacedAdoption' not in dfs:
+            return {}
+        else:
+            episodes = dfs['Episodes']
+            placedAdoptions = dfs['PlacedAdoption']
+            
+            placedAdoptions['DATE_PLACED'] = pd.to_datetime(placedAdoptions['DATE_PLACED'], format='%d/%m/%Y',errors='coerce')
```
dezog: I took this line out, as we aren't actually using this value for any date comparisons, or checking that it's formatted correctly - only making sure it's filled in. 

Now this rule won't trigger for incorrectly formatted dates, due to `to_datetime` replacing them with NaN's that get picked up later by the `episodes_with_errors = merged[merged['DATE_PLACED'].isna()]` line. I also added a test case to make sure of this.

validator903/validators.py
```diff
@@ -1,6 +1,40 @@
...
+
+            episodes = episodes.reset_index()
+
+            place_codes = ['A3','A4','A5','A6']
+
+            placeEpisodes = episodes[episodes['PLACE'].isin (place_codes)]
+
+            merged = placeEpisodes.reset_index().merge(placedAdoptions, how='left', on='CHILD', ).set_index('index')
```
dezog: I deleted the `reset_index` from this line as you've already done this above. This second time weirdly creates an extra column called 'level_0' which I think in this case would contain the same values as 'index', and do no harm particularly but isnt needed

tests/test_validators.py
```diff
@@ -1,6 +1,24 @@
...
 
+def test_validate_551():
+    fake_data_episodes = pd.DataFrame({
+        'CHILD': ['0','A','B','C','D','E','F','G','H','I'],
+        'PLACE': ['x','A3', 'A4', 'A5','A6','A3','A4','x','A3','A5'],
+    })
+    fake_data_placed_adoption = pd.DataFrame({
+        'CHILD': ['A','B','C','D','E','F','G','H'],
+        'DATE_PLACED': [pd.NA,pd.NA,pd.NA,pd.NA,'01/01/2020','15/04/2020',pd.NA,'28th Jan 1930'],
+    })
```
dezog, line 12: Added some extra test cases - 

added '0' at the start to make sure the re-indexing stuff was working right
'G' to make sure we're not picking up episodes with the wrong PLACE code
'H' to make sure we're not picking up incorrectly formatted dates
and 'I' to make sure we are picking up children missing from the PlacedAdoption table entirely

dezog (approved):
Great work on this, tricky rule - made a few changes which I've explained in comments here, let me know if you have any questions about those.

PR #400, opened 2021-11-14
Added rule for issue 241 (error 558)

Steve-Far:
This is very similar to 551 so used that as template, please review..

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -1,6 +1,38 @@
 import pandas as pd
 from .types import ErrorDefinition
 
+def validate_558():
+    error = ErrorDefinition(
+        code='558',
+        description='If a child has been adopted, then the decision to place them for adoption has not been disrupted and the date of the decision that a child should no longer be placed for adoption should be left blank. if the REC code is either E11 or E12 then the DATE PLACED CEASED date should not be provided',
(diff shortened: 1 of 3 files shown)


validator903/validators.py
```diff
@@ -1,6 +1,38 @@
...
+
+            placeEpisodes = episodes[episodes['REC'].isin(rec_codes)]
+
+            merged = placeEpisodes.merge(placedAdoptions, how='left', on='CHILD').set_index('index')
+
+            episodes_with_errors = merged[merged['DATE_PLACED_CEASED'].notna()]
+
+            error_mask = episodes.index.isin(episodes_with_errors.index)
+
+            error_locations = episodes.index[error_mask]
```
dezog, line 30: This bit's doing a bit of a loop, turning the index numbers from `episodes_with_errors` into a mask on the rows of `episodes`, then retrieving those same numbers from that index. 

You could just return: `episodes_with_errors.index.to_list()`

dezog (approved):
Yep this looks great, nice one!

validator903/validators.py
dezog: Doesn't make a difference logically so probably not worth changing right now. Merging

PR #401, opened 2021-11-15
Added rule for issues 232 (error 547)

Steve-Far:
Hi Desi can you review please. It's working as is **but** if I add a column of data with the 'HEALTH' as pd.NA and one of the row values as '1' then it fails and I'm not sure why!


dezog (approved):
I added some extra test cases and it seems to be working just fine! Feel free to submit a pull request with failing tests, that way I can take a look through the logs and see what's going on

PR #405, opened 2021-11-15
Added 377

### LINKED ISSUE #164: Error code 377 - Only two temporary placements coded as being due to holiday of usual foster carer(s) are allowed in any 12-month period.
#### Error code
377
#### Description to display
Only two temporary placements coded as being due to holiday of usual foster carer(s) are allowed in any 12-month period.
Check will fail if there are more than two episodes with a placement code of T3 in any 12- month period.  To qualify as a temporary placement only two holidays of the usual foster carer(s) are allowable in any 12-month period.
#### DfE suggested coding
Count of episodes with `<PL>` = 'T3' must not be more than 2 within a 12-month period

PhilDeakin:
As above

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -3763,3 +3763,33 @@ def validate_529():
             return {'Episodes': epi.index[error_mask].to_list()}
 
     return error, _validate
+
+def validate_377():
+    error = ErrorDefinition(
+        code='377',
+        description='Only two temporary placements coded as being due to holiday of usual foster carer(s) are ' +
+                    'allowed in any 12- month period.',
+        affected_fields=['PLACE'],
+    )
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        else:
+            epi = dfs['Episodes']
+            epi['DECOM'] = pd.to_datetime(epi['DECOM'], format='%d/%m/%Y', errors='coerce')
+            epi.reset_index(inplace=True)
+
+            # RNE for T3 must be P or B  (See rule 380)
+            potent_cohort = epi[epi['RNE'].isin(['P', 'B']) & (epi['PLACE'] == 'T3')]
+
+            # Here I'm after the CHILD ids where there are more than 2 T3 placements.
+            count_them = potent_cohort.groupby('CHILD')['CHILD'].count().to_frame(name='cc')
+            count_them.reset_index(inplace=True)
+            count_them = count_them[count_them['cc'] > 2]
+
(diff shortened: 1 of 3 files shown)


validator903/validators.py
```diff
@@ -3762,4 +3762,34 @@ def _validate(dfs):
...
+    def _validate(dfs):
+        if 'Episodes' not in dfs:
+            return {}
+        else:
+            epi = dfs['Episodes']
+            epi['DECOM'] = pd.to_datetime(epi['DECOM'], format='%d/%m/%Y', errors='coerce')
+            epi.reset_index(inplace=True)
+            epi.sort_values(['CHILD', 'DECOM'], inplace=True)
+            # RNE for T3 must be P or B  (See rule 380)
+            potent_cohort = epi[epi['RNE'].isin(['P', 'B']) & (epi['PLACE'] == 'T3')]
```
dezog: I reckonwe should stick to the wording of the rule in this case, so if the LS (380) & the PLACE (377) both need looking at for an episode, that can happen in one pass rather than having to fix 380 before noticing this one.

validator903/validators.py
```diff
@@ -3762,4 +3762,34 @@ def _validate(dfs):
...
+
+            # Here I'm after the CHILD ids where there are more than 2 T3 placements.
+            count_them = potent_cohort.groupby('CHILD')['CHILD'].count().to_frame(name='cc')
+            count_them.reset_index(inplace=True)
+            count_them = count_them[count_them['cc'] > 2]
+
+            err_coh = potent_cohort[potent_cohort['CHILD'].isin(count_them['CHILD'].to_list())]
```
dezog: Currently this will flag all rows for CHILD id's with 3 or more T3 ep's, it'd be better to just flag the T3 rows. I'll add:
 `err_coh = err_coh[err_coh['PLACE'] == 'T3']` after this line

Small point, but you can use `.isin()` on a pandas Series, so you could skip the `to_list()` in this line

dezog (approved):
Nice one Phil, approving this with a couple of tweaks as commented

validator903/validators.py
dezog: Oh, also a bug - this should be matching on the rows of the original episodes so that we get the right indexes out, not the rows of the `potent_cohort`

dezog: Oh, also a bug - this should be matching on the rows of the original episodes so that we get the right indexes out, not the rows of the `potent_cohort`

PR #408, opened 2021-11-16
[RFC] 205A-205D

### LINKED ISSUE #110: Error code 205A (new) - Child identified as UASC last year is no longer UASC this year, but date UASC ceased in both years does not support this
#### Error code
205A (new)
#### Description to display
Child identified as UASC last year is no longer UASC this year, but date UASC ceased in both years does not support this
Child was reported as UASC last year, but not this year, and date UASC status ceased is not in last year.
#### DfE suggested coding
If `<CURRENT_COLLECTION_YEAR>` `<UASC>` = '0' and `<CURRENT_COLLECTION_YEAR>` -1 `<UASC>` = '1' then:  â€¢`<DUC>` in `<CURRENT_COLLECTION_YEAR>` -1 must be  in `<CURRENT_COLLECTION_YEAR>` -1 and â€¢`<CURRENT_COLLECTION_YEAR>` `<DUC>` must be null

### LINKED ISSUE #111: Error code 205B (new) - Child previously identified as UASC is also UASC this year, but date UASC ceased in both years does not support this
#### Error code
205B (new)
#### Description to display
Child previously identified as UASC is also UASC this year, but date UASC ceased in both years does not support this
Child reported as UASC last year is reported as UASC again this year, but: â€¢date UASC ceased last year is not the child's 18th birthday (the default), or â€¢date UASC ceased this year is not either the child's 18th birthday or a date in the current collection year.
#### DfE suggested coding
If `<CURRENT_COLLECTION_YEAR>` `<UASC>` = '1' and `<CURRENT_COLLECTION_YEAR>` -1 `<UASC>` = '1' then: â€¢`<CURRENT_COLLECTION_YEAR>` -1 `<DUC>` must be the child's 18th birthday â€¢`<CURRENT_COLLECTION_YEAR>` `<DUC>` must be in `<CURRENT_COLLECTION_YEAR>` or = `<CURRENT_COLLECTION_YEAR>` -1 `<DUC>`

### LINKED ISSUE #112: Error code 205C (new) - Child not identified as UASC either this year or last year but date UASC ceased has been provided
#### Error code
205C (new)
#### Description to display
Child not identified as UASC either this year or last year but date UASC ceased has been provided
If a child is not UASC then date UASC ceased should not be provided.
#### DfE suggested coding
If `<CURRENT_COLLECTION_YEAR>` and `<CURRENT_COLLECTION_YEAR>` -1 `<UASC>` = '0' then both `<CURRENT_COLLECTION_YEAR>` and `<CURRENT_COLLECTION_YEAR>` -1 `<DUC>` should be Null.

### LINKED ISSUE #113: Error code 205D (new) - Child identified as UASC this year but not identified as UASC status provided for the child last year
#### Error code
205D (new)
#### Description to display
Child identified as UASC this year but not identified as UASC status provided for the child last year
An unaccompanied asylum seeker acquires UASC status on starting to be looked after for the first time. If a child is UASC in the current year then any existing record for the child in the previous year must also show them as being UASC.
#### DfE suggested coding
If `<CURRENT_COLLECTION_YEAR>` `<UASC>` = '1' then `<CURRENT_COLLECTION_YEAR>` -1 `<UASC>` should be '1'

SLornieCYC:
Closes #110
Closes #111 
Closes #112 
Closes #113

This makes use of the new `<UASC>` column added via #538.


SLornieCYC:
Okay, now I've got stuck @dezog.

Problem: validation rule definition is based on the value of the `<UASC>` field (1 or 0), **however** that field only exists in the xml file specification - and **not** in the csv specification (instead you only include children who are UASC).  I assume the DFE portal programmatically adds the 1/0 flag to datasets loaded from csv based on whether or not the child existed in the UASC file (although it doesn't _actually_ say that in the technical specification).

For this tool we're working the other way around, loading uploaded xml file data into the csv table format, which I imagine is dropping the 1 or 0 value from the `<UASC>` field.  This means that although I have been able to create equivalent logic for rules 205A and 205B (but not exactly following the spec), the same is not true for rules 205C and 205D as the required information does not exist at all.   

I think we probably need to amend config.py and ingress.py (and wherever else??) so that we either a) actually hold `<UASC>` and populate with the xml value (and derive the value for csv imports) and create the rules properly or b) drop records from an xml import where `<UASC>` is 0 and continue trying to approximate logic for all four rules.

dezog:
Hi @SLornieCYC, you've hit the nail on the head here. I think option a), reading the UASC value from .xml and inferring it for .csv would be ideal. There are some other derived field

However, given where we're at time-wise, I think for now we should go with the quick'n'dirty option of simply basing it off who has rows in UASC table, but adding a note to the `ErrorDefinition.description` for rules involving the UASC field to highlight this to users. I'll add an issue to address this properly in future. What do you think of....

> "NOTE: If you uploaded a .xml dataset, the actual value of the 'UASC' field in the Header table is currently ignored! This value is instead assumed to be 1 there is an entry for the child in the UASC table, and 0 otherwise."

dezog:
Actually, having a closer look at the definitions I see what you mean, 205C would make no sense to include as things are (or ever, for csv uploads...). 

205D I think we could fudge as "if the child is in UASC table this year, and they are found in last year's Header table, then they must also be found in last year's UASC table"?

SLornieCYC:
Hi @dezog, I think (!) I've managed to cover all bases here: branching logic to handle the existance of `<UASC>` field (future) and equivalent logic to evaluate the data in it's absence (current).  Tests written for what I think should and shouldn't be valid under each methodology.

tests/test_validators.py
```diff
@@ -103,11 +103,11 @@ def test_validate_205B_noUASCcol():
     result = error_func(fake_dfs_205_noUASCcol)
     assert result == {'UASC':[0,1]}
 
-# Temporary until can handle UASC field properly.
-def test_validate_205C_noUASCcol():
-    error_defn, error_func = validate_205C()
-    result = error_func(fake_dfs_205_noUASCcol)
-    assert result == {'UASC':[]}
+### Temporary until can handle UASC field properly.
```
SLornieCYC: Although you've commented out this test, and the 205C rule in validators.py, there's still the "normal" 205C test above this at line 76.  One version with the `<UASC>` column (from xml format) and one version to test without (current csv format).  The rule in validators.py was written to return nothing in the latter case while the tool isn't recording and populating the `<UASC>` column (other rules have close approximations of the "right" outcomes).

SLornieCYC:
Once #538 has been implemented I'll revisit this PR to get rid of all the work-around stuff and link up with the new `<UASC>` column (I think I assumed that would go in the UASC dataframe rather than Header so will need some changes here).

dezog:
Thanks Stephen, that sounds good. I've merged that now. 

I also added an entry 'file_format' to dfs['metadata'] which will be either 'csv' or 'xml', which might be useful to work around the fact that the UASC column will be missing from dfs['Header'] if csv's not including UASC.csv are loaded

SLornieCYC:
Okay.  All the updated rules appear to work independently (pytest for the specific test_validate_x) but fail the full set of tests. First two failures are because this branch was created before the UASC column was added in ingress.py - not sure about the last two failures.

dezog:
Great stuff Stephen, I'll merge in from main and have a look

dezog:
So, in test_integration the test csv's dont go through the read_csv_from_text function; they're read in directly and dont get the UASC column added to header.  

Going to have to update the test functions to account for this

tab1tha (approved):
This is great work !.

PR #428, opened 2021-11-19
Added 576

### LINKED ISSUE #255: Error code 576 - There is an open missing/away from placement without authorisation period in last year's return and there is no corresponding period recorded at the start of this year.
#### Error code
576
#### Description to display
There is an open missing/away from placement without authorisation period in last year's return and there is no corresponding period recorded at the start of this year.
The start date of the open missing/away from placement without authorisation period in last year's return must match the start date of the first missing/away from placement without authorisation period in this year's return.
#### DfE suggested coding
If `<MIS_START>` >`= `<DECOM>` and less than or = `<DEC>` where episode `<DEC>` is not Null then `<MIS_END>` must be provided

PhilDeakin:
Added 576


dezog (commented):
This is an interesting rule in that the error doesn't necessarily correspond to an existing field in a row of the relevant table, so it doesn't fit neatly into our current way of locating errors to the user.

If a child has an open episode at the end of last year's data, but:

- they have a missing episode this year with a different start date
  - that row might not be a mistake, but a valid, separate episode
- they have a missing episode with no start date filled in
  - seems likely to be the error, but again, it might be a separate episode
- there's no rows for that child in this year's missing table 
  - this is definitely an error, but where do we list it?

What do you think of the idea of attributing the error in the header table instead for errors like this?

PhilDeakin:
Hmmm.... I suppose although the row in this year's missing file probably isn't an error, it's the line of interest in that we would look at it and in all likelihood enter the same missing start/end details on the previous year's missing file as there is sometimes a delay in recording them.
I think it's better that the missing file row is highlighted rather than the header file because the header tells us nothing about the missing dates.

dezog:
Ah I see, yeah that makes sense, though it still leaves the case where they're not in this year's Missing table at all. I'll add some logic inside the check to flag only those children the Episodes table, if that sounds good to you?

dezog (approved):

PR #429, opened 2021-11-19
Added 580

### LINKED ISSUE #259: Error code 580 (new) - Child is missing when ceased being looked after but reason episode ceased not 'E8'.
#### Error code
580 (new)
#### Description to display
Child is missing when ceased being looked after but reason episode ceased not 'E8'.
If a child is missing when they cease to be looked after on their 18th birthday, then reason episode ceased should be 'E8' (Other).
#### DfE suggested coding
If `<MISSING>` = 'M' and `<MIS_END>` = DEC and the DOB + 18 years, then `<REC>` must equal 'E8'.

PhilDeakin:
As above


PhilDeakin:
As the only valid missing codes are M and A, I'm not sure why they've made reference to them being M or A in the coding description.

dezog:
Looks like only 'M' should count as an error, 

> As the only valid missing codes are M and A, I'm not sure why they've made reference to them being M or A in the coding description.

Sounds like only 'M' should count as an error, I've changed it to include this - let me know if I'm missing something

image

dezog (approved):
Added `MISSING == 'M'`, otherwise all good!

PR #439, opened 2021-11-22
Added 442

### LINKED ISSUE #200: Error code 442 - Unique Pupil Number (UPN) field is not completed.
#### Error code
442
#### Description to display
Unique Pupil Number (UPN) field is not completed.
An UPN or UPN unknown code is required for all children looked after other than those looked after under an agreed series of short- term breaks (legal status of V3 or V4).
#### DfE suggested coding
Where any episode present, with an `<LS>` not = 'V3' or 'V4' then `<UPN>` must be provided

tab1tha:
Added 442

Closes #200


validator903/validators.py
```diff
@@ -1,6 +1,33 @@
...
+      episodes.reset_index(inplace=True)
+      header.reset_index(inplace=True)
+
+      code_list = ['V3', 'V4']
+
+      # merge left on episodes to get all children for which episodes have been recorded even if they do not exist on the header.
+      merged = episodes.merge(header, on=['CHILD'], how='left', suffixes=['_eps', '_er'])
+      # Where any episode present, with an <LS> not = 'V3' or 'V4' then <UPN> must be provided
+      mask = (~merged['LS'].isin(code_list)) & merged['UPN'].isna()
+      episode_error_locs = merged.loc[mask, 'index_eps']
```
dezog, line 26: This is failing because it's returning [nan, 4] instead of [0, 4] for the header errors, since some rows of `index_er` will have missing values due to the left merge. 

Another consideration is that children with multiple episodes will currently get an error on their header entry for every episode. 

The easiest thing would be to do .dropna and .unique on the `header_error_locs`

dezog (changes requested):
Hey Tambe, please see comment below

validator903/validators.py
tab1tha: Done. 
I have also pushed a last commit that makes the .dropna() unnecessary. The nan was raised because I mistakenly put the CHILD ID as NaN instead of the variable itself. Thank you.

dezog: The dropna will still be necessary, as someone might upload data with episodes for a child who is not in the Header -- those rows in `merge` will end up giving us nan's in the Episodes errors

dezog: I've changed this and merged

dezog (approved):
Added the dropna, otherwise great!

PR #443, opened 2021-11-23
89 validate 1002

### LINKED ISSUE #89: Error code 1002 - This child has no previous episodes of care, therefore should not have care leaver information recorded.
#### Error code
1002
#### Description to display
This child has no previous episodes of care, therefore should not have care leaver information recorded.
This young person has no previous episodes of care recorded. There should not be any information recorded for them on their 17th, 18th, 19th, 20th or 21st birthday, that is, whether they are in touch, and activity and accommodation.
#### DfE suggested coding
If count of episodes = '0' then `<IN_TOUCH>`; `<ACTIV>` and `<ACCOM>` should not be provided

SLornieCYC:
See note on commit and in-line in validators.py regarding longitudinal issues with this rule. Closes #89


dezog (approved):
Nice one, Stephen. All good, I just tweaked it a little so it will run with only the current year's episodes if that's all there is

SLornieCYC:
Only thing to be aware of there is loading current year only will produce even more false-positives.  The only children who would pass the test in that case (out of the entire care leaver cohort aged 17-21) are those who turned 17 or 18 during the year and left care on or before their birthday.  Honestly I'm still very much in two minds about whether this rule is worth including in the tool at all given the limited data we have available to test.

> For young people who turned 17 or 18 during the latest year, care leaver information is required if they left care before or on their birthday (for example, if a young person turned 17 or 18 in the latest year and left care before their 17th or 18th birthday then care leaver information is required; or if a young person left care on their 17th or 18th birthday in the latest year, then care leaver information also is required). However, if the young person is still in care on their 17th or 18th birthday, then care leaver information is not required. For example, if a young person turned 17 or 18 in the year, and was in care on their 17th or 18th birthday, but left care the day after, then we do not require care leaver information for this young person.

dezog:
Very much agree with you on this. I think what I'll do is just leave this PR open for now, and either merge it when we add something like a 'deselect iffy rules' button, or list it as 'not included' and close it. 

Going forwards I'd say let's make a note of these sorts of rules but leave them until last to code

PR #460, opened 2021-11-26
Added 435

### LINKED ISSUE #195: Error code 435 - Reason for new episode is that child's placement has changed but not the legal status, but this is not reflected in the episode data recorded.
#### Error code
435
#### Description to display
Reason for new episode is that child's placement has changed but not the legal status, but this is not reflected in the episode data recorded.
Where the reason for new episode is coded P or T the legal status of the current episode must be the same as the previous one and the placement must have changed (in most cases at least one of placement postcode, placement URN, placement type or placement provider will have changed).  Check the legal status and placement codes in current and previous episodes. If correct as recorded the reason for new episode code should be amended.  This applies from the 2016 to 2017 data collection onwards.
#### DfE suggested coding
Where `<RNE>` = 'P' or 'T' then `<LS>` must = `<LS>` of previous episode and at least one of `<PL>`, `<PL_POSTCODE>`, `<URN>` or `<PLACE_PROVIDER>` must be different

PhilDeakin:
Added 435... pytest is very slow now.


dezog:
> Added 435... pytest is very slow now.

just made a change that should speed up the tests a bit (on new branches), let me know how things compare to previously!

dezog (approved):
Looks good!

PR #463, opened 2021-11-29
[WIP] Added validator for rule 1001.

### LINKED ISSUE #88: Error code 1001 - The episodes recorded for this young person suggest they are not a relevant or a former relevant child and therefore should not have care leaver information completed.
#### Error code
1001
#### Description to display
The episodes recorded for this young person suggest they are not a relevant or a former relevant child and therefore should not have care leaver information completed.
This young person has not been looked after for at least 13 weeks after the age of 14 (these may be different episodes) with at least some care after the age of 16 (see children looked after collection guide for details) and ceased to be looked after through adoption.  There should not be any information recorded for them on their 17th, 18th, 19th, 20th or 21st birthday, that is, whether they are in touch, and activity and accommodation.
#### DfE suggested coding
If `<IN_TOUCH>`; `<ACTIV>` or `<ACCOM>` are provided THEN total days of previous episodes where `<LS>` is not = 'V3' or 'V4' must be >= to 91 days between the `<DOB>` + 14 years and the `<DEC>` of the final episode of care, with at least 1 day of that total between the `<DOB>` + 16 years  and the `<DEC>` of the final episode, and the final episode `<REC>` must not = 'E11' or 'E12'

SLornieCYC:
*Need to add more test cases for pass and failure (including cases with missing data, date gaps etc).
*Currently reports an issue in test_integration.py which I don't understand.
*This is another one which will return lots of false-positives due to lack of historic data (especially for older care leavers who left care more than one year previously - i.e. those aged 19-21).

Closes #88

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -2,39 +2,79 @@ from validator903.validators import *
 import pandas as pd
 
 def test_validate_1001():
-    fake_episodes_prev = pd.DataFrame([
-      {'CHILD': '101', 'LS': 'C2', 'DECOM': '15/06/2016', 'DEC': '20/12/2020', 'REC': 'E17'},#Pass
-      {'CHILD': '102', 'LS': 'C1', 'DECOM': '08/10/2017', 'DEC': '03/03/2018', 'REC': 'E4a'},#Fail
-      {'CHILD': '102', 'LS': 'V4', 'DECOM': '06/03/2018', 'DEC': '12/08/2020', 'REC': 'E4a'},#Fail
-      {'CHILD': '103', 'LS': 'C1', 'DECOM': '11/05/2015', 'DEC': '07/04/2019', 'REC': 'E12'},#Ignore
-      {'CHILD': '104', 'LS': 'C1', 'DECOM': '26/11/2017', 'DEC': '19/07/2020', 'REC': 'X1'},#Pass
+    # DOB always 01/01/2000
+    # next to each episode, approx days in each age bracket is listed like so:      under 14  :  14-16  :  over 16
+    eps = pd.DataFrame([
+        # [0] - PASS: more than 91 relevant days
+        {'CHILD': '1111', 'LS': 'C2', 'DECOM': '01/01/2014', 'DEC': '01/02/2014', 'REC': 'o'},  # :30:
+        {'CHILD': '1111', 'LS': 'C2', 'DECOM': '01/02/2015', 'DEC': '01/06/2015', 'REC': 'o'},  # :120:
+        {'CHILD': '1111', 'LS': 'C2', 'DECOM': '01/01/2016', 'DEC': '01/02/2016', 'REC': 'o'},  # ::1
+
+        # [1] - FAIL: less than 91 days
+        {'CHILD': '2222', 'LS': 'C2', 'DECOM': '01/01/2010', 'DEC': '01/02/2014', 'REC': 'o'},  # :30:
+        {'CHILD': '2222', 'LS': 'C2', 'DECOM': '01/01/2010', 'DEC': pd.NA, 'REC': 'o'},  # Duplicate DECOM, missing
+                                                                                         # DEC - should get dropped
+        {'CHILD': '2222', 'LS': 'C2', 'DECOM': '25/12/2015', 'DEC': '01/02/2016', 'REC': 'o'},  # :7:30
+
+        # [2] - FAIL: more than 91 days but not after 14th bday
+        {'CHILD': '3333', 'LS': 'C2', 'DECOM': '01/01/2010', 'DEC': '01/03/2014', 'REC': 'o'},  # 120:60:
+        {'CHILD': '3333', 'LS': 'C2', 'DECOM': '25/12/2015', 'DEC': '04/01/2016', 'REC': 'o'},  # :7:3
+
+        # [3] - PASS: more than 91 days, all after 16th bday
+        {'CHILD': '4444', 'LS': 'C2', 'DECOM': '01/01/2016', 'DEC': '01/02/2016', 'REC': 'o'},  # ::30
+        {'CHILD': '4444', 'LS': 'C2', 'DECOM': '25/12/2016', 'DEC': '01/04/2017', 'REC': 'o'},  # ::120
+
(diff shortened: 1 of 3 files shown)


SLornieCYC:
This one was tough to get my head around!  I'm sure there must be simpler and shorter ways to get the outcome.

dezog:
Yikes, proper mind-scrambler of a description on this one - good job! Taking a look now

by the way, as we're now getting very short on time til the big unveiling, I'm thinking it'd be best to prioritise rules that won't have problems with missing past data from until after the launch. The same amount of time spent on ones that definitely work would probably be more impactful this week

tests/test_validators.py
```diff
@@ -1,6 +1,41 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_1001():
+    fake_episodes_prev = pd.DataFrame([
+      {'CHILD': '101', 'LS': 'C2', 'DECOM': '15/06/2016', 'DEC': '20/12/2020', 'REC': 'E17'},#Pass
```
dezog: You're right, more test episodes needed here. It's a shame these are so fiddly and time-consuming to get set up right.

It might be easiest to start by nicking a test episodes table from another similar test (maybe one of 363-367?)

dezog: You seem to be heading in the right direction here though :+1:

validator903/validators.py
```diff
@@ -3,6 +3,88 @@
...
+
+
+            # Work out how long child has been in care since 14th and 16th birthdays.
+            episodes_merged = episodes.reset_index().merge(header[['CHILD', 'DOB', 'DOB14', 'DOB16']], how='left', on=['CHILD'], suffixes=('','_header'), indicator=True).set_index('index')
+
+            v3v4_ls = episodes_merged['LS'].str.upper().isin(['V3','V4'])
+            index_v3v4_ls = episodes_merged[v3v4_ls].index
+            episodes_merged.drop(index_v3v4_ls, inplace=True)
+
+            episodes_merged['DECOM14'] = episodes_merged[["DECOM", "DOB14"]].max(axis=1)
```
dezog, line 1527: This is what's causing problems at the moment, though I'm not 100% sure exactly why.

It seems to be down to missing/invalid dates slipping through:
`ValueError: cannot convert float NaN to integer'

Dropping NaN's whether or not episodes_last is present, and including the DOB in that, might fix it

dezog (commented):
Think I've figured out why it's failing at the moment, please see the inline comment on the validator, although as I mentioned above, I reckon it'd be worth waiting to sort this one out until either launch or we've got all the rules that we're sure will work right done. (...but whatever keeps you coding! 😀)

validator903/validators.py
```diff
@@ -1221,54 +1221,72 @@ def _validate(dfs):
...
+            episodes_merged = (episodes
+                               .reset_index()
+                               .merge(header[['CHILD', 'DOB', 'DOB14', 'DOB16']],
+                                      how='inner', on=['CHILD'], suffixes=('', '_header'), indicator=True)
+                               .set_index('index'))
+            episodes_merged['DURATION'] = (episodes_merged['DEC'] - episodes_merged['DECOM']).dt.days
+            v3v4_ls = episodes_merged['LS'].str.upper().isin(['V3', 'V4'])
+            episodes_merged['DURATION V3/V4'] = episodes_merged['DURATION'] * v3v4_ls.astype(int)
+
+            v3v4_ls = (v3v4_ls
```
SLornieCYC: I don't think this bit is required? Rule 1001 only cares about whether the episode is `<LS>` == V3/V4, not whether the episode also meets duration restrictions of that legal status.  

EDIT: Meaning the bit starting with line 1283 - not quite sure how to get the snip to show the bit I want!

validator903/validators.py
```diff
@@ -1277,25 +1295,32 @@ def _validate(dfs):
...
+            # print(episodes_merged[['CHILD', 'DOB', 'LS', 'REC', 'EVER_ADOPTED', 'DURATION V3/V4']])
 
+            has_care_after_14 = oc3['TOTAL14'] >= 91
+            has_care_after_16 = oc3['TOTAL16'] >= 1
 
-            #Work out final cohort
+            valid_care_leaver = has_care_after_14 & has_care_after_16
 
-            error_mask = ended_adopted | ~valid_care_leaver
+            # Find out if child has been adopted
```
SLornieCYC, line 1548: Isn't this duplicating the work on 'EVER_ADOPTED' at line 1300-1302 above?

EDIT: Meaning the bit starting with line 1317 here - not quite sure how to get the snip to show the bit I want!

validator903/validators.py
dezog: So I was basing that off this bit in the guidance:

image

However after a bit more digging, it sounds like (a) that would only apply only to V3, and not to V4 episodes; and (b) since those would be forbidden by other validation checks anyway (365, 367), you're right there's no need to include them here.

dezog: Unless it is actually possible for a child to end up with illegal V3 episodes which require bypassing 365 or 367? I guess it depends on:
- Whether the legal status in reality changes when that happens, so those would end up being something other than V3 on the return
-  if they did stay V3, whether an LA would actually include them in working out who to collect OC3 data for, or just bypass this rule too

Feels like we'd be very much talking edge cases here though - pretty sure we can just discount V3/V4 episodes here without causing anyone major problems come return time!

validator903/validators.py
dezog: Good spot - yep! looks like the 'EVER_ADOPTED' column doesnt get used for anything, my bad!

I think i was unsure whether only excluding children whose last episode is listed as E11/E12 from this error or excluding children with any E11/E12 episode makes more sense, but as above I think it's largely academic, as it ought to be the last episode.

dezog: btw you should be able to select a range by click-dragging on the line number column. it might take a few goes though as it seems a bit janky

dezog (approved):
Made the V3/V4 changes you notes, and I added a check to exclude anyone over the age of 20 from the rule, since they'd presumably have have left care at least 2 years ago, so they'd definitely be false positives.

PR #479, opened 2021-12-01
Adds workflow publish to PyPI and removes tests on Python 3.9 that run very slow

kws:
* Adds GitHub workflow to build and deploy to PyPI every time a release is created. Remember to bump the project version name before releasing. See README.
* Python builds on 3.9 are very slow due to no compiled wheel for Pandas - we could cache this, but it would be slow every time a dependency is updated. Instead test on 3.8 which, albeit not our target version, should be safe enough.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,37 @@
+name: Publish Python Package
+
+on:
+  release:
+    types: [published]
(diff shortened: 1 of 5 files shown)


dezog (approved):
Thanks for this - I'll try deploying a new release later today 🤞

PR #482, opened 2021-12-01
Added 392B

### LINKED ISSUE #465: Error code 392b - Child is looked after but no postcodes are recorded. [Added 2020 v1.1]
**Note: This one appears to have been missed from the git issues list!**

#### Error code
392b
#### Description to display
Child is looked after but no postcodes are recorded.
Postcodes should be completed if child is looked after unless the legal status is V3 or V4 or the child is UASC, or was formerly UASC
#### DfE suggested coding
If `<LS>` not = ‘V3’ or ‘V4’ and `<UASC>` = ‘0’ and `<COLLECTION YEAR>` -1 `<UASC>` = ‘0’ and `<COLLECTION YEAR>` -2 `<UASC>` = ‘0’ then `<HOME_POST>` and `<PL_POST>` should be provided

For further info, refer to the DfE validation document [here](https://www.gov.uk/government/publications/children-looked-after-return-2021-to-2022-technical-specifications)

PhilDeakin:
As above


dezog (approved):
nice one Phil, I just changed so it runs without UASC tables present, as thatll only affect a few children in many LA's anyway, but added to the error message that there may be FP's for UASC children

PR #488, opened 2021-12-01
Log failing errors to file and continue

dezog:
Hey @kws, I've made a start on logging error checks that fail to run, but thought it'd be best to get your OK if you get the chance:

- Added a handler to the logger to create a logfile named like '903 Validator -- 01-12-2021 21.35.19.log'
  - Not sure if this necessary, or the best way to get from here to a 'Download Logfile' button - would it be better to set this up on the pyodide end? And/or send the logs over there as text instead of a file?
- Wrapped the error function calls in Validator in a try/except, and logged the exceptions
- Validator now collects lists of the errors that failed to run or were skipped due to missing tables. Nothing happens to these yet, though


kws (approved):
I haven't run it, but assuming it works, the changes look sensible

PR #493, opened 2021-12-03
Added validator for rule 199.

### LINKED ISSUE #64: Error code 199 - Episode information shows child has been previously adopted from care.
#### Error code
199
#### Description to display
Episode information shows child has been previously adopted from care.
For confidentiality reasons, if a looked after child has been previously adopted from care (reason episode ceased codes E11 or E12), and subsequently becomes looked after again, then further episode information cannot be added for the child under the same record.  Check that the reason previous episode ceased is correct. If the child was previously adopted from care, then record the post adoption episodes under a new child id code and with a new UPN. The two records should remain separate and unlinked.
#### DfE suggested coding
If any episode has a `<REC>` = 'E11' or 'E12' then no later episode should be provided for this `<CHILDID>`

SLornieCYC:
Closes #64


dezog (approved):
Finally getting down to these last few! I've reworked this one to flag CHILD on all episodes after the first adoption episode, I hope that makes sense

SLornieCYC:
Wow, this one really was a lot simpler than I was trying to make it!

dezog:
Don't worry, it always turns out there was a simpler way in hindsight!

PR #503, opened 2021-12-06
Added 577

### LINKED ISSUE #256: Error code 577 - Child ceased to be looked after but there is a missing/away from placement without authorisation period without an end date.
#### Error code
577
#### Description to display
Child ceased to be looked after but there is a missing/away from placement without authorisation period without an end date.
Please enter an end date for the missing/away from placement without authorisation period to match the date the child ceased to be looked after.
#### DfE suggested coding
If `<REC>` not Null or 'X1' and `<MIS_START>` not Null, then `<MIS_END>` must be provided and must =`<DEC>`

tab1tha:
Added 577

Closes #256

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -1,6 +1,39 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_577():
+  fake_data_episodes = pd.DataFrame([
+      {'CHILD': '101', 'DEC': '08/03/2020', 'REC': 'E45',},  # 0 pass
+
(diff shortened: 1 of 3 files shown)


dezog (approved):
made a few changes - better to use `.isna` instead of checking for `pd.NA` as it might be a different type of NaN (i think read_csv generates numpy float type nan's?), and i think the '=' in the rule definition should be a '>' as presumably MIS_END before DEC is fine. also, please use 4-space indentation in future to match the rest of the code

PR #505, opened 2021-12-07
Added 543

### LINKED ISSUE #228: Error code 543 - Child is aged 10 or over at 31 March and has been looked after continuously for 12 months yet conviction information has not been completed.
#### Error code
543
#### Description to display
Child is aged 10 or over at 31 March and has been looked after continuously for 12 months yet conviction information has not been completed.
Conviction information is required for children aged 10 or over at 31 March of the current collection year who have been looked after continuously for at least 12 months.
#### DfE suggested coding
If `<DOB>` >`= 10 years prior to `<COLLECTION_END_DATE>` and `<CONTINUOUSLY_LOOKED_AFTER> = 'Y' then `<CONVICTED>` should be provided

tab1tha:
Added 543

Closes #228


tests/test_validators.py
```diff
@@ -2,6 +2,59 @@
 import pandas as pd
 
 
+def test_validate_543():
+    fake_data_episodes = pd.DataFrame([
+        #  'CONTINOUSLY_LOOKED_AFTER':['N' ,'Y', 'Y', 'N', 'Y'],
+        # The desired CLA values for this test, outlined in the preceding line, are generated from this data solely based on the number of episodes in the current year (DECOM). The other variables (LS, RNE, DEC) are kept neutral.
```
dezog: sadly you cant take this shortcut as add_CLA_column uses RNE == S, REC != X1, LS and in [V3, V4] along with the DEC & DECOM to tell whether an episode means a child was/wasnt looked after continuously.

i suggest just pinching one of the eps DFs from test_utils.py

tests/test_validators.py
```diff
@@ -2,6 +2,59 @@
...
+        'DOB': ['08/03/2020', '22/06/2000', pd.NA, '13/10/2000', '10/01/2027'],
+        'CONVICTED': [1, pd.NA, 1, 1, pd.NA],
+        # 0 ignore CLA is nan
+        # 1 fail conditions are met but CONVICTED is not provided
+        # 2 ignnore DOB is nan
+        # 3 ignore CLA is not Y
+        # 4 ignore DOB is less than 10 years prior to Collection_end
+    })
+
+    metadata = {'collection_start': '01/04/2031',
```
dezog: as this stands, none of the episodes are during the collection year, so none of the children are continuously in care - the DEC's are all in 2020

validator903/validators.py
```diff
@@ -5,6 +5,35 @@
...
+    )
+
+    def _validate(dfs):
+        if 'OC2' not in dfs or 'Episodes' not in dfs:
+            return {}
+        else:
+            oc2 = dfs['OC2']
+            collection_end = dfs['metadata']['collection_end']
+            # add CLA column
+            oc2 = add_CLA_column(dfs, 'OC2')
```
dezog, line 22: this function adds a column called 'CONTINUOUSLY_LOOKED_AFTER' to the listed dfs which is True/False - see below

validator903/validators.py
```diff
@@ -5,6 +5,35 @@
...
+
+            # to datetime
+            oc2['DOB'] = pd.to_datetime(oc2['DOB'], format='%d/%m/%Y', errors='coerce')
+            collection_end = pd.to_datetime(collection_end, format='%d/%m/%Y', errors='coerce')
+
+            # If <DOB> >= 10 years prior to <COLLECTION_END_DATE>and<CONTINUOUSLY_LOOKED_AFTER> = 'Y' then <CONVICTED> should be provided
+            mask = (collection_end > (oc2['DOB'] + pd.offsets.DateOffset(years=10))) & oc2['CONVICTED'].isna()
```
dezog: you should also include `& oc2['CONTINUOUSLY_LOOKED_AFTER']` in the mask, it should make more sense now it's working correctly!

dezog (changes requested):
hey @tab1tha, looking at this i noticed a bug with the `add_CLA_column` function, which was returning True for children whose care episodes were not in the collection_year. It also now takes into account collection_end so it should work as expected for previous years' data too.

i've merged that into this branch but there's still a few issues with this one, have a look at the inline comments

it looks like errors 545 (#508) and 546 (#509) will need more or less the same changes, so i'll merge main into those too

basically:
- use the test episodes from test_utils.py (you should be able to see which rows are CLA there)
- change the rule to make use of the column added by `add_CLA_col`

dezog (approved):

PR #521, opened 2021-12-13
Fix rule 358

### LINKED ISSUE #518: Bug: Rule 358 - Age calculation incorrect
**Describe the bug**
rule  358 says Child with this legal status should not be under 10, but "There is an error here, the DOB is 2007 and the DECOM 2021 so it's not calculating correctly"

**To Reproduce**
validate an episode with a child over 10 years old and relevant LS

**Proposed fix**
Check date comparison code

SLornieCYC:
Closes #518


dezog (approved):
Great, thanks Stephen

PR #522, opened 2021-12-13
Added 632

### LINKED ISSUE #274: Error code 632 - Date of previous permanence order not a valid value.
#### Error code
632
#### Description to display
Date of previous permanence order not a valid value.
Date should be entered in the format DD/MM/YYYY and should contain a valid date. The date should be before the start of this current episode of care.  Acceptable value where date is not known will include zz, where zz is day, month or year value.
#### DfE suggested coding
If provided `<DATE_PERM>` should be prior to `<DECOM>` and in a valid format and contain a valid date  Format should be DD/MM/YYYY or one or more elements of the date can be replaced by zz if part of the date element is not known.

tab1tha:
Added 632

Closes #274

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -18,9 +18,9 @@ def test_validate_632():
   fake_dfs = {'PrevPerm':fake_data_prevperm, 'Episodes':fake_data_episodes}
   error_defn, error_func = validate_632()
   result = error_func(fake_dfs)
-  assert result == {'Episodes':[2], 'PrevPerm':[1]}
+  # assert result == {'Episodes':[2], 'PrevPerm':[1]}
   # assert result == {'Episodes':[0,2,5], 'PrevPerm':[0,1,4]}
-  # assert result == {'Episodes':[0,2], 'PrevPerm':[0,1]}
+  assert result == {'Episodes':[0,2], 'PrevPerm':[0,1]}
 
 def test_validate_165():
   fake_data_oc3 = pd.DataFrame({
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -20,7 +20,9 @@ def test_validate_632():
   result = error_func(fake_dfs)
   # assert result == {'Episodes':[2], 'PrevPerm':[1]}
   # assert result == {'Episodes':[0,2,5], 'PrevPerm':[0,1,4]}
-  assert result == {'Episodes':[0,2], 'PrevPerm':[0,1]}
+  assert result == {'Episodes':[0,], 'PrevPerm':[0,]}
+  # desired
+  #assert result == {'Episodes':[0,2], 'PrevPerm':[0,1]}
 
 def test_validate_165():
   fake_data_oc3 = pd.DataFrame({
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -3,9 +3,10 @@ import pandas as pd
 
 def test_validate_632():
   fake_data_prevperm = pd.DataFrame({
-      'CHILD': ['101', '102', '103', '104', '105', '106'],
-      'DATE_PERM': ['xx/10/2011', '17/06/2001', '01/05/2000', pd.NA, '05/zz/2020', 'zz/05/2021'],
+      'CHILD': ['101', '102', '103', '104', '105', '106', '107', '108', '109', '110'],
+      'DATE_PERM': ['xx/10/2011', '17/06/2001', '01/05/2000', pd.NA, '05/zz/2020', 'zz/05/2021', 'zz/zz/zzzz', 'zz/zz/1993', 'zz/zz/zz', '01/13/2000' ],
   })
+
   fake_data_episodes = pd.DataFrame([
       {'CHILD': '101', 'DECOM': '20/10/2011', },  # 0 fail DATE_PERM is not in appropriate format
       {'CHILD': '102', 'DECOM': '19/11/2021', },  # 1 pass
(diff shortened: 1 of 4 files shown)


validator903/validators.py
```diff
@@ -4,6 +4,61 @@
...
+      episodes.reset_index(inplace=True)
+      prevperm.reset_index(inplace=True)
+      merged = episodes.merge(prevperm, on='CHILD', how='left', suffixes=['_eps', '_prev'])
+      
+      # check that date is of the right format
+      dte = merged['DATE_PERM']
+      def valid_date(dte):
+        try:
+          # if the date cannot be split on / then its format is inappropriate
+          lst = dte.split("/")
```
dezog: the split method doesnt raise an error if the separator isn't found, it just returns a list of length 1 with the entire string. it'd be better to get rid of this try/except and use len:
```
if len(dte.split("/")) != 3:
    return False
```

In general it's not ideal to wrap everything in a try/except unless you're catching a specific error and know what will cause it. Otherwise it can hide any bugs that take place within the `try` and make it hard to figure out what's up.

validator903/validators.py
```diff
@@ -4,6 +4,61 @@
...
+          # The date should have two (day), two(month) and four(year) elements
+          # if ([].append(len(i) for i in lst) != [2,2,4]):
+            # return False
+          for i in lst:
+            try:
+              int(i)
+              # if element is non-numeric, check if placeholder is appropriate
+            except ValueError:
+              if not 'zz' in i.lower():
```
dezog: this should be `if i.lower() != 'zz':`

otherwise e.g. '19zz' or '2zzbhvv;' would end up counting as valid

I see what you were going for as it looks like the year requires four z's. Unfortunately we'll have to be explicit about it, I think the best we can do is change the loop to: 

```
z_list = ['zz', 'zz', 'zzzz']
for i, zeds in zip(lst, z_list):
    if i != zeds:
        # ...

File: validator903/validators.py [hunk 3]
```diff
@@ -4,6 +4,61 @@
...
+                return False
+              else:
+                return True
```
dezog: don't return True yet, as this will only ever do one pass through the loop - omit this else block

validator903/validators.py
```diff
@@ -4,6 +4,61 @@
...
+        except:
+          return False
+          
+      # convert dates from strings to appropriate format.
+      merged['DECOM'] = pd.to_datetime(merged['DECOM'], format='%d/%m/%Y', errors='coerce')
+      merged['DATE_PERM'] = pd.to_datetime(merged['DATE_PERM'], format='%d/%m/%Y', errors='coerce')
+
+      # If provided <DATE_PERM> should be prior to <DECOM> and in a valid format and contain a valid date Format should be DD/MM/YYYY or one or more elements of the date can be replaced by zz if part of the date element is not known.
+      #mask = valid_date(dte) | (merged['DATE_PERM'] >= merged['DECOM'])
+      mask = merged['DATE_PERM'].apply(valid_date)
```
dezog: the `valid_date` function assumes it's being passed a string (by relying on the .split method) - this will cause an AttributeError, which is currently getting caught by the generic `except`

tests/test_validators.py
```diff
@@ -1,6 +1,25 @@
 from validator903.validators import *
 import pandas as pd
 
+def test_validate_632():
+  fake_data_prevperm = pd.DataFrame({
+      'CHILD': ['101', '102', '103', '104', '105',],
+      'DATE_PERM': ['xx/10/2011', '17/06/2001', '01/05/2000', pd.NA, '05/zz/2020'],
```
dezog: I would add a couple extra checks for DATE_PERM values:
'zz/zz/1993' - valid
'zz/zz/zzzz' - valid
'zz/zz/zz' - fail (I think - wrong zeds)
'01/13/2000' - fail (bad month)
'01/zz/2000' - fail (zeds followed by not-zeds)

dezog (changes requested):
Yikes, this is one of those surprisingly tricky rules, took a lot of figuring out to get my head around how to do it. I've added a few comments in the code, but overall I think the process should be changed a bit to the following, so that it's possible to check partially complete dates like 'zz/zz/2003' against the DECOM:

1. `DATE_PERM_dt = DATE_PERM.apply(valid_date)` (see new version below)
2. drop all episodes except each child's first
3. `mask = ((DATE_PERM_dt > DECOM) | DATE_PERM_dt.isna()) & (DATE_PERM != 'zz/zz/zzzz') `
    - if DATE_PERM_dt is after first DECOM it's an error (should be before)
    - If DATE_PERM_dt is null, it's an error (wrongly formatted date)
    - unless DATE_PERM is 'zz/zz/zzz', then it's valid (cant do date comparison)
---
Instead of returning True/False, `valid_date` now converts the string to a datetime if possible. 'zz/zz/zzzz' results in NaT, so we have to check for that in 3. above
```
def valid_date(dte):
    lst = dte.split('/')
    
    if len(lst) != 3:
        return pd.NaT

    z_list = ['zz', 'zz', 'zzzz']
    # We set the date to the latest possible value to avoid false positives
    offset_list = [pd.DateOffset(months=1, days=-1),
                   pd.DateOffset(years=1, days=-1),
                   None]
    already_found_non_zeds = False
    date_bits = []

    for i, zeds, offset in zip(lst, z_list, offset_list):
        if i == zeds:
            # I'm assuming it is invalid to have a date like '01/zz/zzzz'
            if already_found_non_zeds:
                return pd.NaT
            # Replace day & month zeds with '01' so we can check if the resulting date is valid
            # and set the offset so we can compare the latest corresponding date
            elif i == 'zz':
                i = '01'
                offset_to_use = offset
        else:
            already_found_non_zeds = True
        date_bits.append(i)
    
    as_datetime = pd.to_datetime('/'.join(date_bits), 
                                 format='%d/%m/%Y', errors='coerce')
    try:
        as_datetime += offset_to_use
    except NameError:  # offset_to_use only defined if needed
        pass  
    return as_datetime  

```

validator903/validators.py
tab1tha: here I was considering the possibility of checking for both zz and zzzz (placeholder for year). I see your point though.

dezog (approved):
Everything looks good, great work Tambe. Just made one tweak so it only triggers if DATE_PERM's invalid, but not if it's missing since this is one of those 'if provided...' ones

tab1tha:
> Everything looks good, great work Tambe. Just made one tweak so it only triggers if DATE_PERM's invalid, but not if it's missing since this is one of those 'if provided...' ones

I see. I included that because the NaNs where causing an AttributeError at the .split() line.

PR #523, opened 2021-12-14
Fix IndexError in rules 352, 387, 389.

### LINKED ISSUE #473: Mystery IndexError issue
We've had a report from someone trying to upload an XML file and the tool failing with an IndexError. They say it worked last month, so it seems likely to be related to the recent updates.

Unfortunately, the resulting error message is not massively informative:

> We ran it last month and it provided a helpful output of our errors this year to date, however I’ve just tried running it again today and it displays the following error message:
> 
> image004
>
> I wondered if there was an issue with today’s XML file, so I input the file which ran successfully last month and this one also produces an error:
>
> image005

SLornieCYC:
*Updated tests to include duplicate CHILD in Header.
*Updated error_locations to get unique index values from episodes_merged.

Closes #473


dezog (approved):
That should do it, nice one :+1:

PR #525, opened 2021-12-16
Fix rule 522

### LINKED ISSUE #524: Bug: Rule 522 - Mask inverted
**Describe the bug**
Rule 522 is currently working backwards and flagging cases where `<DATE_PLACED_CEASED>` > `<DATE_PLACED>` rather than where `<DATE_PLACED>` > `<DATE_PLACED_CEASED>`.

**Proposed fix**
Replace > with < in mask expression and update test asserts accordingly.

SLornieCYC:
*Reversed mask to correctly flag end-before-start and updated test asserts.
Closes #524

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -799,16 +799,16 @@ def test_validate_520():
 
 def test_validate_522():
     fake_placed_data = pd.DataFrame([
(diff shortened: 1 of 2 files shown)


dezog (approved):
Good spot, thanks Stephen

PR #528, opened 2021-12-17
Resolves case sensitivity of rules

### LINKED ISSUE #332: Validation rules should not be case sensitive except where specified in the definition
For example:

- [ ] 149 (https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/commit/d2bbfae6cd849032338ce5861e7b6563e1077361) - "E4a" (mixed case as recorded locally) fails the validation but is allowed by the DFE return portal.

- [ ] 392c (https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/commit/cfdfba686abbf11b8e761af68470c6f956934a6d) - lower case postcode fails the validation. I can't evidence this being allowed by the DFE return portal but see no reason why it shouldn't be accepted.

tab1tha:
Capitalizes all string input

Closes #332


tab1tha:
The function recommended in this pull requests does the relevant changes only to the string columns and leaves boolean and numeric columns untouched as shown.
It was also seen to have no effect on strings containing numbers like '26', '1' and so on, as expected.

image

dezog:
Nice one Tambe, this is great, but any rules such as 632 that rely on comparisons to lower case strings will also need updating. Not sure off the top of my head if there are others, but it's worth doing a sweep before we merge this. 

(e.g. changing the zz's to ZZ's in 632 would make it work again)

Beware that the test data in test_validators.py don't go through the ingress functions, and so won't flag this up

tab1tha:
I have gone through all the code in the validators.py file and found that, apart from 632, the other rules are implemented in capital letters. 
The changes you requested have been made in the latest commit.

dezog (approved):
tested, appears to work fine! nice one

PR #529, opened 2021-12-21
Debugs error 602

### LINKED ISSUE #526: Bug: Rule 602 - Rule inverted
**Describe the bug**
Rule is currently flagging all children adopted in year (AD1) with an in-year episode `<REC>` = E11 or E12.  

**Expected behavior**
It should be flagging children in AD1 who **don't** have an in-year episode `<REC>` = E11 or E12.

**Proposed fix**
Either:
- the mask needs reversing to return all episodes that **didn't** end `<REC>` = E11 or E12 in-year (will also need to add .unique() to the production of err_list), **or**
- err_list needs to have a .dropna() where `<REC>` is null (i.e. has no episode ending `<REC>` = E11 or E12).

tab1tha:
This assumes that the bug in error 602 is because the conditions were combined using the & conditional statement and as such, it only works when both conditions are true instead of when either is true.

Closes #526


dezog:
as noted in the thread on #526, i think we need to save the original index by doing .reset_index() first, then .set_index('index') before the .index

tab1tha:
Sidenote: I removed the .unique() in [the commit](https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/pull/529/commits/7fb4f0a3808efe393f1f66077e28ca1af0a65d12) because the merge is left on ad1, hence we are not expecting any duplicates.

dezog (approved):

dezog (approved):

PR #531, opened 2021-12-22
Fixed rule 331.

### LINKED ISSUE #530: Bug: Rule 331 - Mask inverted
**Describe the bug**
Rule 331 is currently working backwards and flagging cases where `<DATE_MATCH>` < `<DECOM>` rather than where `<DATE_MATCH>` > `<DECOM>`.

[Note: The coding description in the guidance document is wrong as it contradicts the message and rule]

**Proposed fix**
Replace < with > in mask expression and update test asserts accordingly.

SLornieCYC:
- Changed < to > in error_mask to correct logic (guidance document coding description is wrong).
- Updated tests to suit.

Closes #530

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -4366,8 +4366,8 @@ def test_validate_331():
 
     result = error_func(fake_dfs)
 
-    assert result == {'AD1': [2, 4, 5],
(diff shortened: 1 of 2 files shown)


dezog (approved):
Good work, thanks Stephen

PR #532, opened 2021-12-22
Read all data as strings to avoid bugs like those affecting rules 181, 202

### LINKED ISSUE #517: Bug: Rule 181 - wrongly flagging null values
**Describe the bug**
@Steve-Far reports rule 181 is flagging null values for children looked after for 12 months

**To Reproduce**
Upload data including null values for any of:

CONVICTED
HEALTH_CHECK
IMMUNISATIONS
TEETH_CHECK
HEALTH_ASSESSMENT
SUBSTANCE MISUSE
INTERVENTION_RECEIVED or
INTERVENTION_OFFERED

**Expected behavior**
Null values should not trigger 181

tab1tha:
Null values were being flagged when required columns were missing in the data. This pull request adds a check that prevents that behaviour.

Closes #517


dezog (commented):
Hey Tambe,

I don't think missing columns is the issue with this one, wouldnt that end up throwing a KeyError at `oc2[fields_of_interest]`?

I suspect this might be down to the same thing as the 'continuously looked after' issues & rule 202, so i think we should look at this in the testing deep dive

dezog (approved):

PR #533, opened 2021-12-23
Change affected_fields on rule 442.

SLornieCYC:
- LS seems unnecessary.


dezog (approved):

dezog:
i agree, good spot, Stephen.

PR #537, opened 2022-01-06
197B-typo

tab1tha:
changes 197B to 197b

This is because in the Dfe document, the latter is used. 197a is already in the right case.


tab1tha:
Hi @dezog I'm closing this since you said it is inherently converted.

PR #539, opened 2022-01-11
[WIP] 534 debug 187

### LINKED ISSUE #534: Bug: Rule 187 - continuously looked after
**Describe the bug**
Children who have not been continuously looked after (e.g. last episode `<DEC>` is before `collection_end`) are being treated as if they were.

**To Reproduce**
- Child has ceasing episode (`<REC>` != X1)  with `<DEC>` before `collection_end`
- Child has either an AD1 or OC3 record with at least one relevant column filled

**Expected behavior**
Children who have not been continuously looked after during the year should be disregarded by this rule.

**Screenshots**

image

image

**Proposed fix**
Don't know! Can't see anything wrong with the python and the test data seems okay, but in the tool front-end I'm getting this for multiple records.

tab1tha:
Children who are not CONTINOUSLY_LOOKED_AFTER should not flag error 187.

Closes #534

### DIFF
--- a/.breakpoints
+++ b/.breakpoints
@@ -4,7 +4,7 @@
       {
         "id": "3426e630-e5e4-41c5-8031-c91c906d2ada",
         "line": 57,
-        "version": 1,
+        "version": 1706,
         "index": 1550
       }
     ]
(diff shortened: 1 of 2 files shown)


tab1tha:
Closing this because 187 has been solved (could by a fix to another issue) and no longer appears to happen. @dezog

PR #553, opened 2022-01-28
Debug 399

### LINKED ISSUE #546: Bug: Rule 399
**Describe the bug**
This rule is triggering on all children with any `<LS>` = V3/V4 episode whereas I think that should only happen when the child does not have any other `<LS>` recorded in-year.

**To Reproduce**
- Upload 'Episodes' and 'Header CSV's, where:
   - 'Header' contains a child with `<MOTHER>` filled in.
   - 'Episodes for same child has one placement with `<LS>` = V4 and one placement with e.g. `<LS>` = V2 (any non-V3/V4 code).
   - Rule 399 is triggered.

**Expected behavior**
Where child has more than one episode in-year that include both `<LS>` = V3/V4 and any other non-V3/V4 `<LS>` this rule should not trigger.  As it currently stands Rule 399 is incompatible with Rule 393.

image

tab1tha:
Ignores cases where **all** LS values are not V3 or V4. 

Closes https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/issues/546


validator903/validators.py
```diff
@@ -1201,6 +1201,13 @@ def _validate(dfs):
             reviews = dfs['Reviews']
 
             code_list = ['V3', 'V4']
+            # Rule adjustment: if child has any other episodes where LS is not V3 or V4, rule should not be triggered. Trigger error 399 only if all child episodes
+
+            # Column that will contain True if LS of the episode is either V3 or V4
+            episodes['LS_CHECKS'] = episodes['LS'].isin(code_list)
+
+            # Column that will contain True only if all LSs, for a child, are either V3 or V4
+            episodes['LS_CHECK'] = episodes.groupby('CHILD')['LS_CHECKS'].transform('max')
```
dezog: This will be True if _any_ of the episodes are V3/V4, you would need `'min'` instead

validator903/validators.py
```diff
@@ -1212,7 +1219,7 @@ def _validate(dfs):
                       .merge(reviews, on='CHILD', how='left'))
```
dezog, line 1221: you could speed this merge up a lot by removing all the unnecessary rows first (those where LS_CHECK is False)

the resulting errors may also contain duplicates where there are multiple reviews, and nan's where a child is not present in reviews or header - make sure all the error lists are .dropna'd and .unique'd

dezog (commented):
You've got the right idea here, couple changes noted below

dezog (approved):
Looks good, nice work!

PR #554, opened 2022-02-01
Amend rule 208: Ignore previously unknown UPNs

### LINKED ISSUE #552: Bug: Rule 208 - Unknown UPN's becoming known shouldn't trigger this error
**Describe the bug**
a helpful LA noticed this error triggering on validated csv's downloaded from the DfE portal, where a child had UN5 code for the UPN field last year's header but it had been filled in in the current year.

They did not expect to see this, as the DfE portal doesn't complain about this error in that case, despite the guidance's claims:

image

@PhilDeakin @bellt101 @Steve-Far @SLornieCYC is anyone able to confirm this?

**Proposed fix**
change so that if unknown UPN code last year but valid UPN this year, it's not an error

tab1tha:
Debugs 208

Closes #552


dezog (approved):
changes look good, i did some further changes to include all the cases with null values

PR #559, opened 2022-02-08
Rule 392B: Update how UASC is referenced.

tab1tha:
Updates UASC mechanism, considering the fact that the UASC column is added during the file ingress process.

### DIFF
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -7027,19 +7027,28 @@ def validate_392B():
     )
 
     def _validate(dfs):
-        if 'Episodes' not in dfs:
+        # comments on similar lines of code are found in validate_406()
+        if 'Episodes' not in dfs or 'Header' not in dfs or 'Header_last' not in dfs:
             return {}
+        elif 'UASC' not in dfs['Header'].columns or 'UASC' not in dfs['Header_last'].columns:
+          return {}
         else:
+            # If <LS> not = 'V3' or 'V4' and <UASC> = '0' and <COLLECTION YEAR> - 1 <UASC> = '0' and <COLLECTION YEAR> - 2 <UASC> = '0' then <HOME_POST> and <PL_POST> should be provided.
             epi = dfs['Episodes']
-            epi['orig_idx'] = epi.index
-            if 'UASC' in dfs:
-                uas = dfs['UASC']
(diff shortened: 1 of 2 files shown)


validator903/validators.py
```diff
@@ -7027,19 +7027,28 @@ def validate_392B():
     )
 
     def _validate(dfs):
-        if 'Episodes' not in dfs:
+        # comments on similar lines of code are found in validate_406()
```
dezog: better to just copy those comments here if they are applicable - saves people having to go hunting them down

validator903/validators.py
```diff
@@ -7027,19 +7027,28 @@ def validate_392B():
...
+        if 'Episodes' not in dfs or 'Header' not in dfs or 'Header_last' not in dfs:
```
dezog: changing to still run when only episodes is provided, as the results would still be useful for the majority of children, as I'd expect false positives from formerly UASC children to be a small fraction for almost all LA's, and that possibility's mentioned in the ErrorDefinition.description

validator903/validators.py
```diff
@@ -7027,19 +7027,28 @@ def validate_392B():
...
             epi['orig_idx'] = epi.index
-            if 'UASC' in dfs:
-                uas = dfs['UASC']
-                err_co = epi.merge(uas, how='left', on='CHILD', indicator=True).query("_merge == 'left_only'")
-                err_co.drop(['_merge'], axis=1, inplace=True)
-            if 'UASC_last' in dfs:
-                uas_l = dfs['UASC_last']
-                err_co = err_co.merge(uas_l, how='left', on='CHILD', indicator=True).query("_merge == 'left_only'")
+            
+            header = header.loc[pd.to_numeric(header['UASC'])==0]
```
dezog: don't forget to add `errors='coerce` to pd.to_numeric calls - otherwise it'll throw an Exception if you get a value that can't be converted

dezog: this isnt strictly necessary any more anyway, as ingress brings everything in as strings now

dezog (approved):
made a few changes, major one being so it runs using whatever tables are present

dezog (approved):

dezog (approved):

PR #576, opened 2022-02-28
Debug 577-578

### LINKED ISSUE #563: Bug: Rule 577 & 578 - Don't work with multiple care periods
**Describe the bug**
If a child has more than one care period during the year (i.e. ended a series of consecutive episodes with REC <> 'X1' and later started another series of episodes with RNE = 'S') these rules currently flag Missing events that are dated within the later care period as in-error because the dates are after the final DEC of the previous care period (ending REC <> 'X1').

**To Reproduce**
Steps to reproduce the problem, e.g.:
- Upload 'Episodes' and 'Missing CSV's, where:
   - 'Episodes' contains a child with one or more episodes covering two distinct care periods.
   - 'Missing' contains a child with one or more missing events that take place after the first care period finished but within the date range of the later care period (which may still be open at collection end!).
- Errors are displayed for the legitimate missing events.

**Expected behavior**
Missing events that exist within a care period (one or more consecutive episodes) should not trigger these errors.

**Screenshots**

image

image

Note that here both Missing events are dated within the second care period but are in-error against the first care period (in this case both care periods present as single episodes, but they could equally exist as multiple episodes across one or more years).

tab1tha:
This pull request closes https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/issues/563


dezog (approved):
Approving, great work on this. I made a few tweaks, let me know what you think:

- to handle PoC's without a final DEC correctly, as the `.max()` on the groupby doesnt work right where there's both nulls and valid values
- removed the check on the REC as well, as I think we can just go off the dates
- removed the exception for V3/V4 episodes as I couldnt see it mentioned in the guidance for Missing episodes
- Also added a couple extra cases to 577 for dealing with ongoing missing eps. 

Not 100% on how we've done these as the wording is a bit vague in places, so do let me know if you disagree with any of this!

Issue #347, opened 2021-10-01
Do we need internal validation rules specific to the tool?


SLornieCYC:
Several data items for a child (`CHILD`, `SEX`, `DOB`) are included in more than one data file.  Where this is the case the values in each file should match all of the other files the child appears in (for csv anyway, not sure how the xml is structured).   In the DFE return portal I'm pretty sure this data is validated on initial file-load (i.e. before the other validations are run) but I don't know whether we actually test for this in the tool right now?

Given several of the validation rules involving `SEX` and `DOB` can be tested using a single file (e.g. age of children in OC3) should we check that the data we have is accurate (well, consistent with other "knowns" at least) - even if it is highly unlikely to ever be different between files.  

It may also be useful to flag up if there are any `CHILD` IDs that exist in one or more files without also being in the main Header file (which should have all children included for the return).  Again I think this is probably something checked by the DFE portal on file-load.

The undocumented "EPI" validation rule from the DFE portal would also fit under this category - which seems to highlight children who exist in the Header file but do not have any in-year Episodes and are not in the OC3 file (error description: "WARNING: Episodes need to be loaded for this child before further validation is possible").

dezog:
Hey @SLornieCYC, this is a good point and it would be well worth doing to prevent any confusing inconsistencies down the line. 

We could write a function to do these checks, and call it in the `run_validation_for_javascript` function in api.py, after the data's read in. If it fails it would then have to notify the user on the javascript side.

Alternately, we could add them as normal validation rules, except with an ErrorDefinition subclass like PreliminaryErrorDefinition or something, which then get handled differently down the line, maybe showing them up at the top in a different colour, or only running the others if these pass. 

Thinking about it, the latter way is probably the better option, as it allows people to see what the actual problems are.

dezog:
Just adding to the list: we should check that all files except Episodes, Missing, and Reviews have only one entry per child

PR #586, opened 2022-03-10
Amend rule 442 (remove duplication in returned errors)

SLornieCYC:
Episodes do not need to be included in the highlighted errors for this rule.  Any child existing in episodes but not in header should be picked up in separate validations (see #347).


dezog (approved):
This makes sense, another nice catch!

PR #587, opened 2022-03-10
Internal validation rules and EPI

### LINKED ISSUE #347: Do we need internal validation rules specific to the tool?

SLornieCYC:
Closes #347 
- Added internal validation rules comparing common data items between files (CHILD, DOB, SEX). 
- Added undocumented EPI validation rule from DFE portal to check whether CHILD in Header also exists in Episodes or OC3.
- Added internal validation rules to check whether CHILD is duplicated in files where CHILD should be unique.


dezog (approved):
Hi Stephen, I'm just back from leave, amazing to see all this in the PR queue! I'm sure having these extra checks is going to be super useful for people.

A couple things worth pointing out:

- We've been calling these 'internal checks' but I think we could do with a more descriptive name for people not familiar with the project's ins & outs? I've gone with 'integrity checks' for now, but happy to hear any other suggestions. 'basic data integrity error', or something along these lines

- I think we should leave out EPI - I think that makes more sense for the DfE portal than ours, mainly because I think we want to allow users to upload any subset of the tables they have at hand, including eg just Header

- I created a IntegrityCheckDefinition subclass of ErrorDefinition to use for these, just to differentiate them, and added separate lists to config.py for validation/integrity checks (still combined into configured_errors with everything). This doesnt really do anything at the moment but should help with handling them differently in the frontend at some stage

SLornieCYC:
Hi Desi, I think "integrity checks" is better (and clearer) than "internal" so happy with that; pretty much does what it says on the tin!

With regards EPI I would tend towards including it in the tool because:
- It makes the tool more consistent with the behaviour LA analysts are familiar with from the actual return.
- Loading single files doesn't make a whole lot of sense for this because there are a lot of cross-dependencies between files.
- If someone _does_ part-load data into the tool they can easily identify and disregard the EPI errors anyway.
- We can add a note to the error description in the tool to clarify what it's saying/doing (i.e. Child in Header but not in Episodes or OC3, or one or more of Episodes/OC3 not loaded).

Anyone using xml source files are highly unlikely to be loading one that only contains header data (composite product usually system-generated), and for csv source files the DFE portal requires Header, Episodes and UASC files to be loaded together as a set - you technically _can't_ load header data there without at least one of the other files involved with EPI anyway (OC3 can be an optional/later load but the logic is unaffected).

dezog:
All good points, I'm convinced - EPI's back! 

I added a note to say the message comes from the DfE portal - feel free to update the wording if needed

PR #588, opened 2022-03-11
Rule 434 - Fixed handling of null values in compared columns.

### LINKED ISSUE #578: Bug: Rule 434 - false positives
**Describe the bug**
this rule is producing errors where there appears to be nothing wrong with the data - LS different; everything else the same

**To Reproduce**
Not sure without access to the data but it was quite common, i'd expect running any actual 903 data through the tool would result in this bug showing up

SLornieCYC:
Closes #578


dezog (approved):
Nice catch. I was worried about the `.fillna('')` for a minute, but it's safe as there's no way an empty string to end up in our df since they get cast to NaN in ingress now

PR #590, opened 2022-04-07
Updated rule 191 to handle when CHILD is missing from OC2 entirely (but is continuously looked after)

### LINKED ISSUE #589: Bug: Rule 191 - OC2 gaps
**Describe the bug**
If a child has been continuously looked after for >=12 months at collection_end they should have OC2 data provided.  If for some reason a child does not appear at all in the OC2 data (e.g. no row with CHILD in csv) - despite having continously looked after status from Episodes data - they are currently not flagged because as coded the rule depends on a record existing in OC2 to highlight.

**To Reproduce**
Steps to reproduce the problem, e.g.:
- Upload 'Episodes' and 'OC2' CSV's, where...
   - 'Episodes' contains a child whose episodes are valid as continuously looked after at collection_end
   - 'OC2' does not contain the child
- Note that error 191 is not flagged on the child despite being continously looked after and without OC2 data

**Expected behavior**
If a child is continously looked after at collection_end they should be flagged as an error in 191 both when they exist in the OC2 data but without all of the mandatory fields, **and** when they do not exist in the OC2 data at all.

**Proposed fix**
Ideally would still flag the OC2 element, but can't see a way to do that under current tool structure.  Instead could maybe change the rule return to be based on the Episodes file rather than the OC2 file.  So that the error only flags once no matter how many episodes in-year make up the continuous status, perhaps also add a filter to only return either earliest/latest starting episode?

SLornieCYC:
Closes #589

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -2973,7 +2973,7 @@ def test_validate_191():
 
     result = error_func(fake_dfs)
 
-    assert result == {'OC2': [1, 6]}
+    assert result == {'OC2': [1, 6], 'Episodes':[3]}
 
 
 def test_validate_516():
--- a/validator903/validators.py
+++ b/validator903/validators.py
@@ -2744,7 +2744,7 @@ def validate_191():
         code='191',
         description="Child has been looked after continuously for at least 12 months at 31 March but one or more "
                     "data items relating to children looked after for 12 months have been left blank.",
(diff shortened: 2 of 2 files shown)


dezog (approved):
Great spot again Stephen, thanks!

Keen to think about how we display errors not associated with a specific row for the upcoming CIN Census tool.

Issue #591, opened 2022-04-12
Bug: Rule 1007 - CLA Portal does not match DFE guidance


SLornieCYC:
**Describe the bug**
The behaviour of this rule in the DFE CLA portal does not appear to match the coding description in DFE's validation rule guidance document (and which has been used to code this rule in this D2I validation tool).  This means that means although we're following the DFE guidance, errors flagged here are different to those that LA users will find in their actual CLA returns.

**Expected behavior**
From observation of records in the DFE CLA portal the rule appears to function more in line with the validation rule description/message, being related to whether the child (aged 17-18) was still in care on the day after their in-year birthday.  I have raised a service request with DFE asking for clarification of the correct coding description - will report back if I hear anything more!

SLornieCYC:
Response from DFE below.

> I think it most straightforward to interpret this rule by reference to the children looked after data collection guide (version 1.3).  The guide states that 
>  
> "For young people who turned 17 or 18 during the latest year, care leaver information is required if they left care before or on their birthday (for example, if a young person turned 17 or 18 in the latest year and left care before their 17th or 18th birthday then care leaver information is required; or if a young person left care on their 17th or 18th birthday in the latest year, then care leaver information also is required).  However, if the young person is still in care on their 17th or 18th birthday, then care leaver information is not required.  For example, if a young person turned 17 or 18 in the year, and was in care on their 17th or 18th birthday, but left care the day after, then we do not require care leaver information for this young person"
>  
> This is the guiding principle and confirms the following statements:
>  
> •	date of birth is between 1 April 2004 and 31 March 2005;
> •	We would expect that a child who doesn't leave care until after their 17th or 18th birthday, would trigger 1007 if the care leaver details are completed; and 
> •	Conversely, care leaver information should be provided if the child is still in care on their 17th or 18th birthday.

dezog:
Hey @SLornieCYC, I've created a [pull request ](https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/pull/596)to fix this, would you mind taking a look to see if it agrees with your understanding of the DfE explanation?

dezog:
By the way, thanks so much for the legwork you've done on all these issues, incredibly helpful for us and for users of the tool!

tab1tha:
> By the way, thanks so much for the legwork you've done on all these issues, incredibly helpful for us and for users of the tool!

I agree.
Thank you @SLornieCYC  !

[...]

PR #596, opened 2022-04-29
(1007) Check if in care on last birthday, instead of at collection end

dezog:
Changed to highlight 17-18 year olds with OC3 data who were looked after on their last birthday.


dezog:
fixes #591

validator903/validators.py
```diff
@@ -3333,32 +3334,42 @@ def _validate(dfs):
...
             # That is, check that 17<=age<19
-            check_dec_rec = merged['REC'].isna() | merged['DEC'].isna()
+
+            merged['bday'] = merged['DOB'].apply(lambda x: x.replace(year=this_year))
+
+            merged.loc[merged['bday'] > collection_end, 'bday'] -= pd.DateOffset(years=1)
+
+            in_care_on_bday = (
+                (merged['DECOM'] <= merged['bday'])
+                & (merged['DEC'] >= merged['bday'])
```
SLornieCYC: Think this is the only section I'd query.

1) Should this be `>` rather than `>=` for DEC comparison?  If child left care on their birthday we still want the OC3 data.

2) What does this do with episodes that are still open at year end? (i.e. DEC is na)?  If there's no DEC child is still in care.

SLornieCYC: Sorry @dezog just realised this one is still sat here. Is there something I should have clicked to "end" my review?

dezog: Ah, yeah it still says 'review required' - after you add the inline comments you have to scroll back to the button at the top and hit Submit Review 
image

My bad though, I should have realised

PR #602, opened 2022-07-12
Updated rule and test on 521 to reverse <= with > in date logic.

### LINKED ISSUE #597: Bug: Rule 521 - Date logic inverted
**Describe the bug**
Rule currently tests for `(merged['DATE_INT'] <= merged['DECOM'])` whereas in-error records should be those where `<DATE_INT>` is **after** `<DECOM>` of applicable episodes.

**Proposed fix**
Replace `<=` with `>` in the above expression (dates are also allowed to be the same ).
Update the tests accordingly.

SLornieCYC:
Closes #597


dezog (approved):
nice catch! looks like a typo in the guidance. i remember being extremely by confused by this...

PR #603, opened 2022-07-12
599 fix 344

### LINKED ISSUE #599: Bug: Rule 344 - Should test for 0 not null
**Describe the bug**
Rule currently tests for `oc3['ACTIV'].notna() | oc3['ACCOM'].notna()` when `<IN_TOUCH>` == 'RHOM' or 'DIED' however per coding rule comments for the validation rule (and the SSDA903 guidance document) the correct answer in these cases is a value of 0 rather than null.

**Proposed fix**
Amend rule to test for value of 0 in `<ACTIV>` and `<ACCOM>` and update tests accordingly.

SLornieCYC:
Closes #599

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -2369,15 +2369,15 @@ def test_validate_442():
 
 def test_validate_344():
     fake_data_oc3 = pd.DataFrame({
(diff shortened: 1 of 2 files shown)


dezog (approved):
you're absolutely right, nice one!

PR #605, opened 2022-07-22
Updated Rule 186 to cover SDQ_REASON

### LINKED ISSUE #598: Bug: Rule 186 - Rule should account for SDQ_REASON
**Describe the bug**
Rule checks whether `<SDQ_SCORE>` has a value and flags those without as in-error, however some children may have `<SDQ_REASON>` populated instead of `<SDQ_SCORE>`.  These children are flagged in-error despite having the alternative field populated.

Although `<SDQ_REASON>` is not included in the coding description for this validation rule it **is** included in the coding rule comments.   Rules 188 and 189 have similar purposes and include `<SDQ_REASON>` in both coding description and comments.

**Proposed fix**
Amend rule (and tests) so that a child without `<SDQ_SCORE>` only flags an error if `<SDQ_REASON>` is also blank.

SLornieCYC:
Closes #598 

- Make rule consistent with behaviour of rules 188 and 189.
- Update test accordingly.

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -2775,11 +2775,13 @@ def test_validate_186():
     ])
 
     oc2 = pd.DataFrame({'CHILD': ['999999', '1', '2', '3',
-                                  '9999', '8', '5555'],
+                                  '9999', '8', '5555','789'],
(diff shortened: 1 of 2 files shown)


dezog (approved):
nice catch, this makes perfect sense 👍

PR #609, opened 2022-08-31
Updated Childrens Home ingress

### LINKED ISSUE #606: Bug Report: Children's Home Ingress does not work with Ofsted files from April 2022
### Problem
Starting in April 2022 Ofsted has dropped the "Settings and Inspection Info" and "Address Details" tabs from the monthly Children's Homes List for LAs file in favour of the "Provider information" tab (introduced previously) which contains merged data from those tabs.  Additionally the "Provider information" tab has been moved from fourth place in the file to first place ("Settings and Inspection info" was previously the first tab with "Address Details" in third place).

This means that `construct_provider_info_table` in **ingress.py** fails when attempting to read from this file because it is trying to access the third sheet in a workbook that now only contains two sheets.

image
image

https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta-validator/blob/0ae95409bc2ebad8a54739e4a5ef51f95e685d11/validator903/ingress.py#L173

### Relevant sheet columns 
The columns in "Provider information" are exactly the same as "Settings and Inspection Info" except that the additional columns from "Address Details" have been inserted between Assurance Visit Start Date and Owner Type.

**Settings and Inspection Info**
image

**Address Details**
image

**Provider information**
image

### Options

1. Change `construct_provider_info_table` to use the "Provider information" tab only (referenced by name rather than sheet index number).  **Pros:** The steps to read and merge from "Address Details" can be deleted from the function as all of the information is available in the single tab.  **Cons:** The function will only work for recent Ofsted Childrens Homes files that include the new merged tab (unsure when this was introduced but believe was during the 2021/2022 return year).
2. Change `construct_provider_info_table` to use the "Provider information" tab by default (referenced by name rather than sheet index number) but fall back on "Settings and Inspection Info" and "Address Details" if that is not available.  **Pros:** The function should work for both new files as well as those that predate the "Provider information" tab existing.  **Cons:** The function requires additional complexity and skip-logic to test for and then deal with the existence of different tab combinations.

SLornieCYC:
Closes #606


tab1tha:
Hi Stephen. This is good work. 
We can set up a session this week or early next week so that I show you how to test it on the files you've got and see if this works !.

When do you prefer? 12-3pm tomorrow or any time after 12noon on Monday works for me.

Before then, it'll be nice to get an idea about your enviroment/tools in case they've changed.

- Are you coding fully on replit or are you able to code offline too (an IDE installed on your computer)?
- Have you ever successfully set up the frontend ([instructions](https://github.com/SocialFinanceDigitalLabs/quality-lac-data-beta)). No worries if you haven't, we'll sort it out.

tab1tha:
Hi @SLornieCYC,
Do check the above message, I forgot to tag you earlier.

SLornieCYC:
Thanks @tab1tha, 

Monday afternoon works for me.

* At the moment I am coding fully on Replit.  I don't currently have access to a locally installed IDE (something for the future maybe).
* I've tried setting up the front end via Replit but haven't managed to get that working yet.  I think I'm missing something with the connection to the python wheel I've imported from the validator repository.

tab1tha:
I just realized that I do not have your email address (no need to put it publicly here).
Could you send me a meeting invite instead @SLornieCYC ?

Email address: tambe.tabitha@socialfinance.org.uk
Time: 3-4 pm on Monday the 5th of September

scripts/check_ch_scp_files.py
```diff
@@ -0,0 +1,4 @@
+#Used when testing amended ingress function for amended file formats.
+from validator903.ingress import construct_provider_info_table
+
+print(construct_provider_info_table('./CH2.xlsx', './SCP2.xlsx').columns)
```
tab1tha: This file was only used for checking that your changes work. 
If you want to remember the code, it'll always exist here in this thread. However, it is not a necessary part of the tool so I'm taking it out.

tab1tha (approved):
These changes have been seen to work on the new and old format of files as tested by @SLornieCYC.
I can also confirm that these changes do not break the tool when files that existed before the government update, are uploaded.

Good work @SLornieCYC !

tab1tha:
Awesome.

tab1tha:
@SLornieCYC 
I formatted it by doing `black ingress.py` in the console after I had installed black. 
You can[ install black](https://pypi.org/project/black/) by running `pip install black` if you want to do it too.

PR #621, opened 2022-11-29
rule 558

### LINKED ISSUE #614: Error code 558 [Amended 2022 v1.1]
#### Error code
558
#### Description to display
Child has been adopted during the year but the date of decision that they should no longer be placed for adoption has been completed.
If a child has been adopted, then the decision to place them for adoption has not been disrupted and the date of the decision that a child should no longer be placed for adoption should be left blank.
#### DfE suggested coding
If latest episode `<REC>` = 'E11' or 'E12' then **if number of `<DATE_PLACED_CEASED> in the year is > 1`** then **latest** `<DATE_PLACED_CEASED>` ~~should not be provided~~ **must be null**

WillLP-code:
updates 558 to fit 2022 v1.1 spec

closes #614


SLornieCYC:
Hi Will, not sure if you've finished amending this rule or not - if so I think you've missed the main substance of the change:

- if number of `<DATE_PLACED_CEASED>` in the year is > 1

validator903/validators.py
```diff
@@ -4362,8 +4362,18 @@ def _validate(dfs):
...
 
+            # List of children who have null dates to be excluded (assumes the null is their most recent)
+            children_with_null = merged[(merged['DATE_PLACED_CEASED'].isna())]['CHILD'].tolist()
+
+            episodes_not_null = merged[~(merged['CHILD'].isin(children_with_null))]
+
+            episodes_not_null['COUNT'] = episodes_not_null.groupby(['CHILD', 'YEAR'], group_keys=False)['DATE_PLACED_CEASED'].transform('count')
+
+            episodes_with_errors = episodes_not_null[episodes_not_null['COUNT'] > 1]
+
```
SLornieCYC: I know this PR was created before the great refactoring (which I think has now finished?) and needs updating to bring into line with the new file structure before it can be merged.  However:

I don't think the `['YEAR']` column and group by is required here.  When the rule states "within the year" this usually means the entire return year **dataset** rather than a specific calendar year (or even only dates within the financial year represented by the return).

At the moment Episode 5 in the test data passes the test because the two `DATE_PLACED_CEASED` dates are in different years (2020 and 1930) but in reality this should fail because both have dates and there is no third record with a NULL date.

lac_validator/rules/lac2022_23/rule_558.py
```diff
@@ -25,7 +25,23 @@ def validate(dfs):
...
+            "CHILD"
+        ].tolist()
+
+        episodes_not_null = merged[~(merged["CHILD"].isin(children_with_null))]
+
+        episodes_not_null["COUNT"] = episodes_not_null.groupby(
+            ["CHILD", "YEAR"], group_keys=False
+        )["DATE_PLACED_CEASED"].transform("count")
+
+        episodes_with_errors = episodes_not_null[episodes_not_null["COUNT"] > 1]
```
SLornieCYC: Re-adding comment post-merge to restructured repo.

I don't think the `['YEAR']` column and group by is required here. When the rule states "within the year" this usually means the entire return year dataset rather than a specific calendar year (or even only dates within the financial year represented by the return).

At the moment Episode 5 in the test data passes the test because the two `DATE_PLACED_CEASED` dates are in different years (2020 and 1930) but in reality this should fail because both have dates and there is no third record with a NULL date.

SLornieCYC:
Testing this one gives a warning:

image

WillLP-code:
> Testing this one gives a warning:
> 
> image

Intersting, I'll have a look now!

WillLP-code:
@SLornieCYC Fixed the warning,  just made a new DF rather than editing the current one.

SLornieCYC (approved):

PR #627, opened 2023-02-14
bugfix 523

WillLP-code:
This PR probably needs to be split into multiple PRs, so I'm leaving it as a draft.

Adds devcontainer and dockerfile to force Python 3.9 in Codespaces (you'll still need to manually pip install poetry and poetry install, I commented out most of the docker because I was getting container build issues,  which I'll sort when I'm doing it in the real PR). Numpy and Pandas take a long time to install, should be fixable by updating to a more recent pip and poetry,

Adds the offline debug file used to run and investigate individual rules (might keep this in a shell form in a real PR?)

Adds a placed for adoption set of test data with errors so 523 has something to do. Also, repeats a child with different data as this is needed to replicate the bug.

When offline debug is run without changing the merge suffixes in 523, it recreates the bug seen in production where certain indexes weren't found in the data, when the merge suffixes are fixed, it then fixes the bug and correctly identifies indexes.


SLornieCYC:
I've updated the other bugfix PRs to the new file structure etc but I don't know anything about docker so am going to skip this one!

PR #639, opened 2023-05-19
Update run-tests.yml to python3.9 for validator integration

tab1tha:
cin_validator runs on python3.9 so lac is being updated to match it.


tab1tha:
Continued here https://github.com/data-to-insight/quality-lac-data-beta-validator/pull/645

PR #641, opened 2023-05-26
Fix rule 452, 453, EPI, 1002, 1010

franciscobmacedo:


tab1tha (approved):
all good. 
Considering that the function you changed was named 452_453_503G_503H, I would've assumed that all the rules in its name will be affected. apparently not.

I do like the idea of fixtures !.

tab1tha:
This fails because of the other rules that we are in the process of resolving. 
Merging this in and will build on it.

Thank you !

PR #648, opened 2023-06-09
manage registry within package of year rules

tab1tha:
**Context of changes**

At first the decorator attached to validate rules [automatically added them](https://github.com/data-to-insight/quality-lac-data-beta-validator/blob/f1354183037ec4a224af9d60130f045db4d60713/lac_validator/rule_engine/__registry.py#L127) to registry whenever a file containing a validate function was opened. 
This caused conflicts because when the files needed to be opened twice (e.g in the case of [loading and running pytest](https://github.com/data-to-insight/quality-lac-data-beta-validator/blob/f1354183037ec4a224af9d60130f045db4d60713/lac_validator/__main__.py#L48)) the registry flagged all rules as duplicates.
As such, [this line](https://github.com/data-to-insight/quality-lac-data-beta-validator/blob/f1354183037ec4a224af9d60130f045db4d60713/lac_validator/rules/lac2022_23/__init__.py#L7) had to be commented to test the rules (`python -m lac_validator test`) and then uncommented to run the rules (`python -m lac_validator run tests\\fake_data\placed_for_adoption_errors.csv tests\\fake_data\\ad1.csv`)
The changes in this pull request fix that so that toggling is no longer necessary. It does this by [attaching a tag](https://github.com/data-to-insight/quality-lac-data-beta-validator/blob/0ed8ba1998b8087b68c951d56a74009c7157cfdf/lac_validator/rule_engine/__registry.py#L134), the ".rule" attribute in this case, that validator rules can be recognised by so that they can be filtered out and added to the registry only when needed.

**Handover TODOs**
- [ ] Rename file lac_validator_class to `lac_validator.py`
- [ ] Drag files that are being imported from validator903 into the lac_validator folder. Allow vs_code to auto-update the imports  so that they now point to lac_validator.
- [ ] Use new registry format to define relationship between years when new rulesets are created. all the TODOs about init and registry refer to this. [This commit ](https://github.com/data-to-insight/quality-lac-data-beta-validator/pull/648/commits/0ed8ba1998b8087b68c951d56a74009c7157cfdf)contains annotations of where the code might need to be changed.

**Notes about the repo.**
- _Day to day changes:_ treat the `feat/code-refactor` branch as the parent branch of all changes. Create branches from it and make pull requests to it. 
- Keeping track of state of code in frontend: Build the wheel (`poetry-build`) from the `feat/code-refactor` branch whenever the frontend needs to be updated. Then merge the `feat/code-refactor` branch into its parent (`refactor/migration`) so that day-to-day merging can keep going on on the `feat/code-refactor` branch while knowing that the `refactor/migration` branch is a reflection of the code being run  in the frontend. This aids in debugging python errors that can arise in the frontend.
- The priority is getting things done. So if that means deleting or totally rewriting large blocks of logic, absolutely do that. I'm not attached to this lines of code. Delete away.

### DIFF
--- a/lac_validator/__main__.py
+++ b/lac_validator/__main__.py
@@ -12,7 +12,6 @@ from lac_validator.utils import process_uploaded_files
 from validator903.ingress import read_from_text
 from validator903.report import Report
 
-
 @click.group()
 def cli():
     pass
--- a/lac_validator/__main__.py
+++ b/lac_validator/__main__.py
@@ -6,7 +6,6 @@ import os
 from pathlib import Path
 
 import lac_validator.lac_validator_class as lac_class
-from lac_validator.ruleset import create_registry
 import importlib
 from lac_validator.utils import process_uploaded_files
 
@@ -33,8 +32,9 @@ def list_cmd(ruleset):
 
     :return cli output: list of rules in validation year.
     """
-    registry = create_registry(ruleset=ruleset)
-    for rule in registry:
+    module = importlib.import_module(f"lac_validator.rules.{ruleset}")
+    ruleset_registry = getattr(module, "registry")
+    for rule in ruleset_registry:
         click.echo(f"{rule.code}\t{rule.message}")
 
 
@@ -118,7 +118,7 @@ def run_all(p4a_path, ad1_path, ruleset, select):
 
     # the rest of the metadata is added in read_from_text() when instantiating Validator
     metadata = {"collectionYear": "2022", "localAuthority": "E09000027"}
-    module = importlib.import_module("lac_validator.rules.lac2022_23")
+    module = importlib.import_module(f"lac_validator.rules.{ruleset}")
     ruleset_registry = getattr(module, "registry")
 
     v = lac_class.LacValidator(metadata=metadata, files=files_list, registry=ruleset_registry, selected_rules=None)
--- a/lac_validator/__main__.py
+++ b/lac_validator/__main__.py
@@ -7,6 +7,7 @@ from pathlib import Path
 
 import lac_validator.lac_validator_class as lac_class
 from lac_validator.ruleset import create_registry
+import importlib
 from lac_validator.utils import process_uploaded_files
 
 from validator903.ingress import read_from_text
@@ -117,13 +118,13 @@ def run_all(p4a_path, ad1_path, ruleset, select):
 
     # the rest of the metadata is added in read_from_text() when instantiating Validator
     metadata = {"collectionYear": "2022", "localAuthority": "E09000027"}
+    module = importlib.import_module("lac_validator.rules.lac2022_23")
+    ruleset_registry = getattr(module, "registry")
 
-    v = lac_class.LacValidator(
-        metadata=metadata, files=files_list, ruleset=ruleset, selected_rules=None
-    )
+    v = lac_class.LacValidator(metadata=metadata, files=files_list, registry=ruleset_registry, selected_rules=None)
     results = v.ds_results
 
-    r = Report(results)
+    r = Report(results, ruleset=ruleset_registry)
     print(f"*****************Report******************")
     print(r.report)
     print(f"*****************Error report******************")
--- a/lac_validator/lac_validator_class.py
+++ b/lac_validator/lac_validator_class.py
@@ -9,7 +9,7 @@ from validator903.datastore import copy_datastore, create_datastore
 
 logger = logging.getLogger(__name__)
 
-# TODO rename this file
+# TODO rename this file so that it doesn't have class in its name.
 class LacValidator:
     """
     Central location for running rules on files.
--- a/lac_validator/lac_validator_class.py
+++ b/lac_validator/lac_validator_class.py
@@ -1,5 +1,4 @@
 import logging
-import datetime
 import pandas as pd
 from typing import Any
 from pandas import DataFrame
@@ -7,8 +6,6 @@ from pandas import DataFrame
 from validator903.types import UploadedFile
 from validator903.ingress import read_from_text
 from validator903.datastore import copy_datastore, create_datastore
-
-from lac_validator.ruleset import create_registry
 
 logger = logging.getLogger(__name__)
 
@@ -22,7 +19,7 @@ class LacValidator:
         self,
         metadata: dict[str, Any],
         files: list[UploadedFile],
-        ruleset,
+        registry,
         selected_rules,
     ):
         self.dfs: dict[str, DataFrame] = {}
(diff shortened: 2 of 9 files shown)


lac_validator/rules/lac2022_23/__init__.py
```diff
@@ -1,7 +1,15 @@
...
\ No newline at end of file
+registry = []
+for rule_file in files:
+    # get all the elements (functions, classes, variables) of the file and their attributes.
+    rule_content = importlib.import_module(f"lac_validator.rules.lac2022_23.{rule_file}")
+    rule_elements = {n:getattr(rule_content, n) for n in dir(rule_content)}
+    # all validator functions have a decorator which attaches a "rule" attribute to them.
+    validator_funcs = [v.rule for k, v in rule_elements.items() if hasattr(v, "rule")]
+    registry.extend(validator_funcs)
+
```
franciscobmacedo, line 21: a few things here:

1 - No need to have so many loops/list comprehension (I count 4 instead of 2 that are really needed).
2 - I don't like variables called "p" or "n" even if they are in a list comprehension: Why not just call them what they are? 
2 I would suggest, for the sake of cleaning the code, for this to be a function, so:

```python
def get_registry():
    for path in Path(__file__).parent.glob("*.py"):
        if path.stem == "__init__":
            continue
        rule_content = importlib.import_module(f"lac_validator.rules.lac2022_23.{path.stem}")
        for name in dir(rule_content):
            element = getattr(rule_content, name)
            if validator_function := getattr(element, "rule", None):
                registry.append(validator_function)
            
    
registry = get_registry()
```

I didn't test this code, so some adaptation might be needed

rpc_main.py
```diff
@@ -24,15 +24,15 @@
...
-def get_rules(ruleset:str="lac2022_23")->list[dict]:
+def get_rules(ruleset:str="lac2022_23")->str:
     """
     :param str ruleset: validation ruleset according to year published.
     :return rules_df: available rule codes and definitions according to chosen ruleset.
     """
-
-    registry = create_registry(ruleset)
+    module = importlib.import_module(f"lac_validator.rules.{ruleset}")
+    ruleset_registry = getattr(module, "registry")
```
franciscobmacedo, line 33: No need to use `getattr`in here I think you can just do `ruleset_registry = module.registry`

lac_validator/rules/lac2022_23/__init__.py
franciscobmacedo: Also, I'm not sure what kaj suggested, but this is quite an expensive piece of code, it takes some time. I wonder if it wouldn't make sense just to call this function on a github action and dump the registry list in a json/yml file? And then you can just load that when you need it? It's far less expensive to load a file than it is to loop through each file every time a user wants to use the tool.

You can make the github workflow run on every merge.

This is a suggestion that you can add to the tasklist if you find it relevant, not at all for you to do know.

franciscobmacedo (approved):
I think it looks ok, but I would strongly advise not to do that registry loop "live", as in, when the users are using the app/CLI. It should run only once a new rule is added and stored somewhere I think

cyramic (commented):
I ran this through `pylint` and `black` modules, and there are a lot of suggested changes I think should be required before a merge so that the code meet things like PEP8 and readability standards.

lac_validator/__main__.py
```diff
@@ -117,13 +118,13 @@ def run_all(p4a_path, ad1_path, ruleset, select):
...
-        metadata=metadata, files=files_list, ruleset=ruleset, selected_rules=None
-    )
+    v = lac_class.LacValidator(metadata=metadata, files=files_list, registry=ruleset_registry, selected_rules=None)
     results = v.ds_results
 
-    r = Report(results)
+    r = Report(results, ruleset=ruleset_registry)
     print(f"*****************Report******************")
     print(r.report)
     print(f"*****************Error report******************")
```
cyramic, line 129: Is there a reason we're outputting to console (print) instead of using a logger that would give more options? Or are these here for more debugging reasons?

lac_validator/lac_validator_class.py
```diff
@@ -8,8 +7,6 @@
 from validator903.ingress import read_from_text
 from validator903.datastore import copy_datastore, create_datastore
 
-from lac_validator.ruleset import create_registry
-
 logger = logging.getLogger(__name__)
 
 # TODO rename this file
```
cyramic: Is this "TODO" tech debt? What is the purpose of this note here?

cyramic: If it is really something that needs to be done, a way to make sure it doesn't get lost is to put it in the issue tracker and not as a comment here where it could potentially be forgotten. Try not to put things like this in code. In the issue tracker, you can link to the file or line that you feel needs a fix in the future.

lac_validator/rules/lac2022_23/__init__.py
cyramic: I second Francisco's suggestion above. I also feel that this code isn't very easily testable as written, is it?  I also wonder if we should put a try/except block around the import_module function to catch if an import fails, but I'm not sure how necessary that is.

rpc_main.py
cyramic: It depends on if that value is potentially nonexistant. If we're sure it always exists, I'd say go with the dot notation, otherwise this way could catch errors more easily.

rpc_main.py
```diff
@@ -76,10 +76,12 @@ def lac_validate(lac_data:dict, file_metadata:dict,  selected_rules: Optional[li
     # lac_data = {"This year":[p4a_path, ad1_path], "Prev year": [p4a_path]}
 
     files_list = process_uploaded_files(lac_data)
+    module = importlib.import_module(f"lac_validator.rules.{ruleset}")
+    ruleset_registry = getattr(module, "registry")
```
cyramic, line 80: These two lines are a repeat of the lines 32,33 above. It might be a good idea to move this to a separate function so you're not repeating yourself and then having to rewrite multiple sections of code if we decide to change in the future.

For instance, if we want to add more error-handling and security checks on the import module step or if there's a bug. Ideally we'd like to do this in one place instead of several.

cyramic (changes requested):

franciscobmacedo:
> I ran this through `pylint` and `black` modules, and there are a lot of suggested changes I think should be required before a merge so that the code meet things like PEP8 and readability standards.

this is not the final PR of the refactoring - that should be another one. Note that the destination branch is not the main one.

lac_validator/lac_validator_class.py
franciscobmacedo: Tambe didn't have the time to do it - so we left it in the code. I agree, we should add a few issues to this, including the tasks in the main comment.

PR #656, opened 2023-07-20
attempt fix 632. run main code with updated actions suite.

tab1tha:


tab1tha:
Github actions fails at the level of installing dependencies.

> Note: This error originates from the build backend, and is likely not a problem with poetry but with numpy (1.17.5) not supporting PEP 517 builds. You can verify this by running 'pip wheel --use-pep517 "numpy (==1.17.5)"'.

Let's update the poetry package versions to match the new tool too and see if it passes.

tab1tha:
So without changing the code, rule_632 [which passed moments ago](https://github.com/data-to-insight/quality-lac-data-beta-validator/pull/655), now fails in github actions.

The priority is not to do any changes to the rule logic, unless as a last resort. 
So we attempt to find package version numbers that will [not raise installation errors](https://github.com/data-to-insight/quality-lac-data-beta-validator/pull/656#issuecomment-1643661867) and will allow rule 632 to pass.

tab1tha:
Although this rule could be made to pass by backdating all package versions and degrading the python version to 3.8, that won't be a good engineering design decision. it locks that project in and is a risky bet to take considering that some of those versions are no longer being maintained (e.g numpy 1.17 can no longer be installed successfully in poetry 2.0's environment).

The way forward has been to find any obvious loopholes in the logic that can be corrected. As such, it was found that the error being raised at the `first_eps_idxs = episodes.groupby('CHILD')['DECOM'].idxmin()` is caused by NaNs in the DECOM field. 

Since the nan DECOMs can be removed without a change to the logic, this was done and rule 632 now passes integration tests.

> To replicate, do
`poetry install` to get recommended package versions.
`poetry shell` to enter virtual env of project.
`pytest` to run full test suite.

Now that this has been shown to work in the old codebase  (validator903), it will be committed to main and the type of change will be implemented in the new codebase (lac_validator)

tab1tha:
validator903 fix: https://github.com/data-to-insight/quality-lac-data-beta-validator/pull/657
lac_validator fix: https://github.com/data-to-insight/quality-lac-data-beta-validator/commit/b6993e7f47e082b6231bfe0ed0dc2afd6082560a also see https://github.com/data-to-insight/quality-lac-data-beta-validator/pull/658

PR #663, opened 2023-08-07
yearly rule versioning

tab1tha:
Log of the changes in this pull request.
- The registry class has been removed and pure dict and list operations are used instead. It's main advantage was that it changed the default iter behaviour of the dict containing the rules such that iter went through values instead of keys. Apart from that, its presence only added avoidable fuzziness so it has been removed.
- The logic of getting validator functions from files and updating them across years has been isolated into standalone functions.
- CI/CD checks were failing because of an unrelated reason (rule 625 couldn't handle NaNs). A separate pull request was created to fix that and the changes have been pulled into this one.


lac_validator/rules/lac2023_24/__init__.py
```diff
@@ -0,0 +1,13 @@
+from lac_validator.rules.lac2022_23 import registry as prev_registry
+from pathlib import Path
+from lac_validator.rule_engine import RuleDefinition
+from lac_validator.rules.ruleset_utils import extract_validator_functions, update_validator_functions
+
+files = Path(__file__).parent.glob("*.py")
+this_year_validator_funcs:dict[str, RuleDefinition] = extract_validator_functions(files)
+# if any rules need to be deleted, add their codes as strings into del_list
+del_list:list[str] = []
+this_year_config = {"deleted": del_list, "added_or_modified": this_year_validator_funcs}
```
franciscobmacedo: I would suggest that the this year config is also a dataclass (or another defined datastructure), since this is only used in the `update_validator_functions` function. That function currently only receives an "anonymous" dict - this would be less error prone and easier to read.

```python3
@dataclass
class YearConfig:
    deleted: list[str]
    added_or_modified: dict[str, RuleDefinition]
```

lac_validator/rules/ruleset_utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+def update_validator_functions(prev_registry, this_year_config:dict) -> dict:
+    """
+    :param dict prev_registry: previous year's registry.
+    :param dict this_year_config: codes of rules that have been added or deleted.
+
+    :return: valid validator functions according to config.
+    :rtype: dict
+    """
+    prev_validator_funcs = prev_registry
```
franciscobmacedo: not sure what's the usefulness of creating this variable?

franciscobmacedo (changes requested):

lac_validator/rules/lac2023_24/__init__.py
tab1tha: You're right.
It is much easier to read if implemented with a dataclass.
Pushing the changes now.

lac_validator/rules/ruleset_utils.py
tab1tha: True. That's redundant. 
I've removed it in the new commit.

PR #668, opened 2023-08-15
Test childrens home/social care providers ingress

WillLP-code:
Adds a test for the ingress of SCP and CH files, pre-empting changing the ingress to take a single file.


lac_validator/__main__.py
```diff
@@ -148,5 +153,12 @@ def xmltocsv(p4a_path):
     click.echo(data_files)
 
 
+@cli.command(name="testscpingress")
+@click.argument("ch_path", required=False)
+@click.argument("scp_path", required=False)
+def cli_test_ingress(ch_path, scp_path):
+    test_construct_provider_info_table()
+
```
SLornieCYC: Not sure what the option arguments are for here as not passed into `test_construct_provider_info_table`.  Also there are no default values set (for when not provided).

tests/conftest.py
```diff
@@ -65,3 +65,16 @@ def dummy_empty_input(dummy_metadata):
...
+
+
+@pytest.fixture(scope="session")
+def dummy_chscp():
+    def read_file(path):
+        with open(path, mode="rb") as f:
+            return f.read()
+
+    fake_ch_dir = os.path.join(os.path.dirname(__file__), "fake_data", "ch_fake.xlsx")
+    fake_scp_dir = os.path.join(os.path.dirname(__file__), "fake_data", "scp_fake.xlsx")
```
SLornieCYC, line 77: File directory for the test files is defined twice; once here and once in test_ingress.py as part of the new `test_construct_provider_info_table` function.  

Perhaps this fixture should just return the two file directory strings with `test_construct_provider_info_table` updated to consume those (and then do the `file_read`)?

tests/test_ingress.py
```diff
@@ -77,3 +78,33 @@ def test_read_xml_from_text():
         assert set(str(dtype) for dtype in val.dtypes) == {
             "object"
         }, f"Got non-objects columns in {name}: \n{val.dtypes}!"
+
+
+def test_construct_provider_info_table(dummy_chscp):
+    """Tests childrens home and social care providers form ingress as both forms
+    and as filepaths to the location of the files."""
```
SLornieCYC, line 90: See comment on `dummy_chscp` in conftest.py.

tests/test_ingress.py
```diff
@@ -77,3 +78,33 @@ def test_read_xml_from_text():
...
+        "PROVIDER_CODES",
+        "REG_END",
+        "POSTCODE",
+        "source",
+        "LA_CODE_INFERRED",
+        "LA_NAME_INFERRED",
+    ]
+
+    scp_path_dir = os.path.join(os.path.dirname(__file__), "fake_data", "scp_fake.xlsx")
+    ch_path_dir = os.path.join(os.path.dirname(__file__), "fake_data", "ch_fake.xlsx")
```
SLornieCYC: See comment on `dummy_chscp` in conftest.py.  This is the second time these test file directories are being defined.

tests/test_ingress.py
```diff
@@ -77,3 +78,33 @@ def test_read_xml_from_text():
...
+    output_from_string = construct_provider_info_table(ch_path_dir, scp_path_dir)
+    string_output_columns = output_from_string.columns.to_list()
+
+    assert string_output_columns == expected_columns
+
+    ch = {}
+    scp = {}
+    ch["file_content"], scp["file_content"] = dummy_chscp
```
SLornieCYC: See comment on `dummy_chscp` in conftest.py. This is where I thought the `read_file` function could happen instead.

SLornieCYC (commented):
I **think** I understand what this is meant to achieve but I couldn't figure out how to pass `dummy_chscp` as a variable to the `testscpingress` cli command when running in Codespace.  That might just be a lack of python knowledge on my part.  Anyway, comments attached!

lac_validator/__main__.py
WillLP-code: Right oops, this was from an old iteration where you could pass it variables, but I ended up hardcoding them. So to run it via the CLI you literally just do python -m lac_validator testscpingress and it'll run. This does assume you local/virtual environment has everything installed, if it doesntyou can poetry install then poetry shell, then it'll work!

tests/conftest.py
WillLP-code: Ahh this is intersting, one generates a bytes file to send to the ingress, and one generates a string, which means it tests the ingress as both a string and a bytes file.

tests/test_ingress.py
WillLP-code: So, I read the file in conftest as that provides a fixture that can be used throughout the pytests if needed, for instance if I need them later or in another function, rather than ONLY having them within the one function.

tests/test_ingress.py
WillLP-code: do you mean, as I want to test both file and string ingress versions, I could return the directory strings from the pytest fixture, as well as the files?

lac_validator/__main__.py
SLornieCYC: That's what I thought I should do, but it gave me this error which led me down the variable rabbithole:

image

tests/test_ingress.py
SLornieCYC: Having read your other comments as well, I think that would be better than my suggestion (moving the read to the test function).  Essentially I was thinking that if we ever need to have more than one version of the dummy scp or ch files the directory reference would only need updating in one place in the code.

lac_validator/__main__.py
WillLP-code: Okay, I understand this now, it's because dummy_chscp is only avaliable udring a pytest. What I'll is remove the CLI to run this test, because there's a CLI to run the pytest itself

WillLP-code:
@SLornieCYC I think I've done everything you've said. I removed the CLI for the ingress test as there's a CLI for just the Pytest, which allows the test to use the fixtures, which is why it wasn't working for you. I've also defined the string just once, as suggested.

lac_validator/__main__.py
```diff
@@ -11,6 +11,11 @@
 from lac_validator.utils import process_uploaded_files
 
 
+from tests.test_ingress import (
+    test_construct_provider_info_table,
+)
+
+
```
SLornieCYC: I don't think this import is needed now the cli command has been removed?

WillLP-code: Done!

SLornieCYC (approved):
All looks good to me!

PR #690, opened 2023-10-05
Fix rule 186 to correctly test columns are empty

### LINKED ISSUE #620: Bug: Rule [186] - [Children aged 4 or over at the start of the year and children aged under 17 at the end of the year and who have been looked after for at least 12 months continuously should have a Strengths and Difficulties (SDQ) score completed.   ]
**Describe the bug**
Whilst checking our data I think this rule is checking whether there is a SDQ Score and SDQ Reason within the OC2 file, where as you dont need a reason if you have a score and dont need a score if you have a reason.  Its currently showing 585 errors within our file, it is correctly discounting the under 4 and over 

**To Reproduce**
Upload a file where OC2 file has A SDQ Score and SDQ Reason this should flag an error, SDQ Score has a value and SDQ Reason is null should not flag an error and vice versa.

**Expected behavior**
No error should be flagged if either of the SDQ Score or SDQ Reason is entered.

**Screenshots**

**Proposed fix (Optional)**
If you have an idea of what might need changing in the code outline that here
The code looks like it is checking the correct values, and that there is no blank in either field, so not sure why this is not working correctly.
& oc2[['SDQ_SCORE', 'SDQ_REASON']].isna().any(axis=1)

I have downloaded the CSV on its own and I now only get errors for Error 188 and 189 (17 errors) and these are based on the fact that there is an SDQ submitted for a 4 year old and under or 17 and over, these are valid errors within the file,

SLornieCYC:
Have also re-written tests. Closes #620

### DIFF
--- a/lac_validator/rules/lac2022_23/rule_186.py
+++ b/lac_validator/rules/lac2022_23/rule_186.py
@@ -38,7 +38,7 @@ def validate(dfs):
         (oc2["4th_bday"] <= collection_start)
         & (oc2["17th_bday"] > collection_end)
         & oc2["CONTINUOUSLY_LOOKED_AFTER"]
-        & oc2[["SDQ_SCORE", "SDQ_REASON"]].isna().any(axis=1)
+        & oc2[["SDQ_SCORE", "SDQ_REASON"]].isna().all(axis=1)
     )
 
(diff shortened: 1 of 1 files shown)


WillLP-code (approved):
Great work, Stephen

PR #711, opened 2023-10-30
Adds ingress for combined SCP and CH lists

WillLP-code:
Adds the ability to check uploads to the FE so the new, combined SCP/CH list can be uploaded.
Assumes people will use what is currently the CH box to upload the combined list (fe changes to come later).
Adds a function to check what's in the upload, adds a function to create the providers list for the metadata for a single list (built on the original providers metadata function) and adds relevant tests.


lac_validator/ingress.py
```diff
@@ -95,10 +95,25 @@ def read_from_text(
...
+            else:
+                raise UploadError(
+                    "Please load both the latest 'Children's Homes' and 'Social Care Providers' lists "
+                    "from Ofsted into their respective boxes above."
+                )
+        else:
+            raise UploadError(
+                "Please load both the latest 'Children's Homes' and 'Social Care Providers' lists "
+                "from Ofsted into their respective boxes above."
+            )
```
SLornieCYC, line 116: ~~Do these two `UploadError` messages need updating to reference the fact that newer lists come in a single file?~~ EDIT: Nevermind, I've seen you're handling that in the new `combined_ch_scp_check` function at line 708!

lac_validator/ingress.py
```diff
@@ -363,6 +378,116 @@ def construct_provider_info_table(CH: UploadedFile, SCP: UploadedFile):
...
+    # if not check whether file includes separate setting and address sheets
+    scpch_df["source"] = "Combined SCPCH List"
+    scpch_df["placement provider code"] = scpch_df[
+        "placement provider code"
+    ].str.replace("/", ",")
+    provider_info_df = scpch_df.rename(
+        columns=dict(zip(scpch_cols, provider_info_cols))
+    )
+    del scpch_df
+    logger.info(f"CH dataframe complete. Creating SCP dataframe {sc.t}")
```
SLornieCYC: Is this long section still required for retrieving data from the new single Ofsted file?  

The try except set-up was added for the two separate files due to an incremental change in one of the templates ([details here](https://github.com/data-to-insight/csc-validator-be-903/issues/606)) but I would hope that for the new file we wouldn't need that? Unless Ofsted have already started fiddling with the structure of the new file!

WillLP-code: You're right, I just didn't know if different LAs had different headings so I pretty much just took the original two file function and re-worked it for the single file, so I'll hop back in and simplify this!

WillLP-code: So I've removed the first try/except, but I've kept the second one because it's an important error to raise

lac_validator/ingress.py
```diff
@@ -363,6 +378,105 @@ def construct_provider_info_table(CH: UploadedFile, SCP: UploadedFile):
...
+        "urn",
+        "local authority",
+        "placement code",
+        "placement provider code",
+        "deregistration date",
+        "setting address postcode",
+    ]
+    logger.info(f"URN lookup bytes recieved. Reading excel files... {sc.t}")
+
+    # check whether file includes consolidated provider information sheet
```
SLornieCYC: This comment not required (not relevant to the new single file and doesn't reflect subsequent code).

lac_validator/ingress.py
```diff
@@ -363,6 +378,105 @@ def construct_provider_info_table(CH: UploadedFile, SCP: UploadedFile):
...
+    scpch_providers = pd.read_excel(scpch_bytes, engine="openpyxl")
+    scpch_providers.columns = scpch_providers.columns.str.lower()
+    logger.debug(
+        f"Reading SCP/CH provider info from excel done. cols:{scpch_providers.columns} {sc.t}"
+    )
+
+    scpch_address = scpch_providers[["urn", "setting address postcode"]]
```
SLornieCYC: Line 441 (defining `scpch_address`) not required (not relevant to the new single file and not actually being used anyway). Edited for clarity,

lac_validator/ingress.py
```diff
@@ -363,6 +378,105 @@ def construct_provider_info_table(CH: UploadedFile, SCP: UploadedFile):
...
+
+    try:
+        scpch_providers = scpch_providers[scpch_cols[:]]
+    except KeyError:
+        raise UploadError(
+            f'Failed to find required columns in Childrens Homes list "Provider information"'
+            f' sheet. Expected: {", ".join(scpch_cols[:])}'
+        )
+    scpch_df = scpch_providers
+    del scpch_providers, scpch_address
```
SLornieCYC: After removing line 441 above (see separate comment) `scpch_address` can be removed here too (from line 451 del statement). Edited for clarity.

lac_validator/ingress.py
```diff
@@ -363,6 +378,105 @@ def construct_provider_info_table(CH: UploadedFile, SCP: UploadedFile):
...
+
+    # if not check whether file includes separate setting and address sheets
```
SLornieCYC: This comment not required (not relevant to the new single file and doesn't reflect subsequent code).

SLornieCYC (changes requested):

lac_validator/ingress.py
SLornieCYC: Great, that looks much neater!

SLornieCYC (approved):

PR #713, opened 2023-10-31
fixes issue with pytest on 503G

WillLP-code:
Pytest for 503G failed because a NaN was being passed to the abs function on 112 of rule_utils that checks placement distances from the dummy dfs in fixtures.

Fixed by putting a distance in and changing the failing dfs in 503 and 453.

This change does not affect the tests for any other rules which check for absolute placement distances (they use different test dfs).

Potentially doesn't resolve issues where an open episode may, for some reason, not have a placement distance and which is passed to this rule, which will cause an error as it'll try to pass it a NaT. One solution could be to slice out open episodes with no distance attached before the abs, then add them to the arror mask afterwards, but I'm not sure I know the 903 well enough to say if that's needed/important? If this rule isn't causing issues in the live version, all I've done is fix the tests so it *should* remain fine in the live version.


SLornieCYC:
Hm. I wonder how that slipped through before. (It's actually one [I wrote originally](https://github.com/data-to-insight/csc-validator-be-903/pull/396) too!).  

think we _should_ capture where an episode is missing the placement distance when the corresponding placement from the other year has one, because that would still be a mismatch in data.

WillLP-code:
> Hm. I wonder how that slipped through before. (It's actually one [I wrote originally](https://github.com/data-to-insight/csc-validator-be-903/pull/396) too!).
> 
> think we _should_ capture where an episode is missing the placement distance when the corresponding placement from the other year has one, because that would still be a mismatch in data.

@SLornieCYC  I have a feeling the way to do this is to update this line so that the .astype()s ignore errors, as so:
`err_mask = (
            abs(merged_co[this_one].astype(float, errors='ignore') - merged_co[pre_one].astype(float, errors='ignore'))
            >= 0.2
        )
`
This means that even if the FE has an empty PL_DISTANCE, it won't break the validator. You're right though, we should still capture where an episode is missing the placement distance when the corresponding placement from the other year has one, I'll think about that now!

WillLP-code:
The push I've just made:
a) allows pd.NAs through the error mask line, without breaking the validator
b) adds instances to it where just one of each of the PL_DISTANCES is a pd.NA to the error mask

SLornieCYC:
That looks right for 503G.  Now I've just got two remaining concerns:

1) I think rules 503G and 453 should both return the same error records from the test dataset.  The coding definition in the DFE documentation **is** slightly different between the two rules which makes 453 narrower in scope, but not in a way that is affected by our test dataset. (453 needs `<RNE>` to be the same in both years as well as having `<DEC>` = NULL).
2) I think either the `compare_placement_coordinates` function in rule_utils.py needs to work for both rules, or the code from that function needs reintegrating into 503G directly (removing the separate function), with both 503G and 453 then separately containing the distance comparison logic (453 would need updating to match the 503G/function logic).

The `compare_placement_coordinates` function was created when the 903 tool was restructured to mirror the cin tool.  Before that the distance comparison behaviour for 503G was included in a single 503x rules function (which has now been abstracted to `field_different_from_previous`) but 453 always had the distance comparison coded separately.

lac_validator/rules/lac2022_23/rule_453.py
```diff
@@ -12,66 +12,59 @@
...
+        this_one = field
+        pre_one = this_one + "_PRE"
+
+        # if subval == 'G':
+        err_mask = (
+            abs(
+                merged_co[this_one].astype(float, errors="ignore")
+                - merged_co[pre_one].astype(float, errors="ignore")
+            )
             >= 0.2
```
SLornieCYC: I think the whole `field = "PL_DISTANCE"` thing at line 20 and then continued here is unnecesary complication (inherited from previous implementation). Wouldn't it be cleaner and more readable to just have `merged_co["PL_DISTANCE"]` and `merged_co["PL_DISTANCE_PRE"]` in-line here?

The `# if subval == 'G':` comment is also legacy from original implementation and no longer needed.

lac_validator/rules/lac2022_23/rule_503G.py
```diff
@@ -11,25 +10,61 @@
...
+        this_one = field
+        pre_one = this_one + "_PRE"
+
+        # if subval == 'G':
+        err_mask = (
+            abs(
+                merged_co[this_one].astype(float, errors="ignore")
+                - merged_co[pre_one].astype(float, errors="ignore")
+            )
+            >= 0.2
```
SLornieCYC: Same as above.  I think the whole `field = "PL_DISTANCE"` thing at line 18 and then continued here is unnecesary complication (inherited from previous implementation). Wouldn't it be cleaner and more readable to just have `merged_co["PL_DISTANCE"]` and `merged_co["PL_DISTANCE_PRE"]` in-line here?

The `# if subval == 'G':` comment is also legacy from original implementation and no longer needed.

SLornieCYC (changes requested):
Left a few notes.  I can't tag the line directly as not in the affected range, but rule_453.py has `import pandas as pd` on line 1 and then duplicated on line 6.

SLornieCYC (approved):
This one looks good to me now!

PR #714, opened 2023-10-31
updated 2023 rule 371 to 16 from 14

### LINKED ISSUE #636: Error code 371 [Amended 2023 v1.1]
#### Error code
371
#### Description to display
Child in semi-independent living accommodation not subject to children’s homes regulations should be at least ~~14~~ **16**.
Where the placement code is H5 the date at the start of the placement should be on or after the child’s ~~14th~~ **16th** birthday.
#### DfE suggested coding
Where `<PL>` = ‘H5’ then `<DOB>` should be >= ~~14~~ **16** years prior to `<DECOM>`

WillLP-code:
closes #636


lac_validator/rules/lac2023_24/rule_371.py
```diff
@@ -0,0 +1,68 @@
...
+    fake_epi = pd.DataFrame(
+        {
+            "CHILD": ["111", "222", "333", "444", "555", "666"],
+            "DECOM": [
+                "01/01/2020",
+                "01/02/2020",
+                "02/03/2020",
+                "15/01/1980",
+                "15/03/2014",
+                "14/03/2014",
```
SLornieCYC: Change these two values from 2014 to 2016 dates so that they continue to correctly test the rule (i.e. now test 16th bday and day before 16th bday, from 14 bday previously).

WillLP-code: I kept trying to fix this and it kept not working, and then it turned out I was changing the test in the 22/23 file not the 23/24 file, but it is now done!

SLornieCYC (approved):

PR #715, opened 2023-11-02
Add rule 217t to 23/24

### LINKED ISSUE #632: Error code 217t [Added 2023 v1.1]
**Note: This rule was added by DFE in the 2023/24 v1.1 validation check document.**

#### Error code
217t
#### Description to display
Placement type 'P2' is not a valid placement type from 28 October 2023
Independent living is not a valid placement from 28 October 2023.
Any independent living placements must have an end date and must have ended on or before 28 October.
These young people may have been placed in regulated supported accommodation. If so then this accommodation should be recorded as 'H5' and the provider URN should be provided.
#### DfE suggested coding
Where `<PL>` = 'P2'; `<DEC>` must be present and must be <= 28/10/2023

SLornieCYC:
Closes #632

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_217t.py
+++ b/lac_validator/rules/lac2023_24/rule_217t.py
@@ -20,11 +20,8 @@ def validate(dfs):
             "28/10/2023", format="%d/%m/%Y", errors="coerce"
         )
 
-        mask = (
-            episodes["PLACE"].isin(["P2"])
-            & (
-                (episodes["DEC"] > max_dec_allowed) | episodes["DEC"].isna()
-            )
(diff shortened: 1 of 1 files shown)


WillLP-code (approved):
This looks good, I just want to check that the failing row on line 42 is absolutely correct? So, are we sure that there isn't another rule that should be flagging that row? But if you're certain, merge away!

SLornieCYC:
Thanks Will; yes line 42 should fail because it's an open episode (no end date) after the date that placement type has ceased to be valid.

SLornieCYC:
I guess this is why `squash merge` is a thing!  Next time I'm gonna use that (or make sure all my commits have sensible titles).

image

PR #718, opened 2023-11-06
rule 582

### LINKED ISSUE #616: Error code 582 [Added 2022 v1.1]

#### Error code
582
#### Description to display
Data is showing that the child was matched and adopted on the same date.
Child is showing as ceasing to be looked after due to adoption, but the date of adoption is the same as the date reported for being matched with adopters.  In most circumstances we expect the date the child is matched with adopters to be before the date of adoption.
#### DfE suggested coding
If `<REC>` = 'E11' or 'E12' then that episode `<DEC>` should not = `<DATE_MATCH>`
Apply for 2023 onwards

WillLP-code:
closes #616 

This implementation of 582 checks that there are no times where, when a child has E11 or E12 in an episode, the DEC for that episode does not match a DATE_MATCH from AD1.

What it does not do is check that the DEC matching the DATE_MATCH from the same episode group. Yet.

### DIFF
--- a/lac_validator/rules/lac2022_23/rule_582.py
+++ b/lac_validator/rules/lac2022_23/rule_582.py
@@ -1,0 +1,81 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="582",
+    message="Child is showing as ceasing to be looked after due to adoption, but the date of adoption is the same as the date reported for being matched with adopters. In most circumstances we expect the date the child is matched with adopters to be before the date of adoption.",
+    affected_fields=["DEC", "DATE_MATCH"],
+)
+def validate(dfs):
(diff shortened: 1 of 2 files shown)


SLornieCYC (approved):
A nice simple rule for a change! Looks good!

lac_validator/rules/lac2022_23/rule_582.py
```diff
@@ -0,0 +1,81 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="582",
+    message="Child is showing as ceasing to be looked after due to adoption, but the date of adoption is the same as the date reported for being matched with adopters. In most circumstances we expect the date the child is matched with adopters to be before the date of adoption.",
```
SLornieCYC, line 8: Usually we've included the shorter error description as that matches what is shown to users in the DFE portal.

PR #719, opened 2023-11-07
rule 583 working

### LINKED ISSUE #617: Error code 583 [Added 2022 v1.1]

#### Error code
583
#### Description to display
More than one date the child should be placed for adoption has been reported, but the earlier decision wasn't revoked
More than one date the child should be placed for adoption has been reported, but an earlier decision has not been revoked.  Check that both dates are required and add in the date the earlier decision was revoked. Note this may be in a previous year.
#### DfE suggested coding
If more than one `<DATE_PLACED>` is present, then all but the latest `<DATE_PLACED>` must have `<DATE_PLACED_CEASED>`
(The latest `<DATE_PLACED>` may or may not have `<DATE_PLACED_CEASED>`)
Apply for 2023 onwards

WillLP-code:
closes #617 

If more than one <DATE_PLACED> is present, then all but the latest <DATE_PLACED> must have <DATE_PLACED_CEASED>
(The latest <DATE_PLACED> may or may not have <DATE_PLACED_CEASED>)

Works on the assumption that this rule coding means that any child with more than one DATE_PLACED_CEASED NaN is a failing row. Does not check if the most recent date placed ceased has a value but an earlier one does not, which now that I submit this PR, I realise it should.

### DIFF
--- a/lac_validator/rules/lac2022_23/rule_583.py
+++ b/lac_validator/rules/lac2022_23/rule_583.py
@@ -15,15 +15,9 @@ def validate(dfs):
     if "PlacedAdoption" not in dfs:
         return {}
     else:
-        # If a child has more than one Na in date_placed_cease, they fail, sending all Na rows as errors
+        # # If a child has more than one Na in date_placed_cease, they fail, sending all Na rows as errors
         df = dfs["PlacedAdoption"]
-        ceased_nans = df[df["DATE_PLACED_CEASED"].isna()]
-        nans_count = ceased_nans["CHILD"].value_counts()
-        too_many_nans_list = list(nans_count[nans_count > 1].index)
-        too_many_nans = ceased_nans[ceased_nans["CHILD"].isin(too_many_nans_list)]
-        too_many_nans_index = list(too_many_nans.index)
 
-        # If a child has an earlier date_placed_ceased Na, they fail even if their total NAs is 1
         df_ordered = df.sort_values("DATE_PLACED", ascending=False)
         df_ordered.drop_duplicates("CHILD", keep="first", inplace=True)
         df_ordered_index = list(df_ordered.index)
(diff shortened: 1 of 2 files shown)


lac_validator/rules/lac2022_23/rule_583.py
```diff
@@ -0,0 +1,96 @@
...
+            {
+                "CHILD": "child2",
+                "DATE_PLACED": "01/02/2020",
+                "DATE_PLACED_CEASED": pd.NA,
+            },  # 2 fail
+            {
+                "CHILD": "child2",
+                "DATE_PLACED": "02/02/2020",
+                "DATE_PLACED_CEASED": pd.NA,
+            },  # 3 fail
```
SLornieCYC: I think `# 3` here should be a pass because that is the latest `<DATE_PLACED>` which is allowed to be blank.

lac_validator/rules/lac2022_23/rule_583.py
```diff
@@ -0,0 +1,96 @@
...
+            {
+                "CHILD": "child4",
+                "DATE_PLACED": "01/02/2020",
+                "DATE_PLACED_CEASED": "02/02/2020",
+            },  # 6 pass
+            {
+                "CHILD": "child4",
+                "DATE_PLACED": "01/01/2020",
+                "DATE_PLACED_CEASED": pd.NA,
+            },  # 7 fail
```
SLornieCYC, line 77: I like that you've tested multiple different outcomes with the same `<DATE_PLACED>` date!

lac_validator/rules/lac2022_23/rule_583.py
WillLP-code: Okay, if anything that's easier to code! Just removed the check for multiple Nas, and now returns only Nas that are not the most recent date placed

SLornieCYC (approved):
Sorry taken a while to get back to look at this!  Looks good!

lac_validator/rules/lac2022_23/rule_583.py
```diff
@@ -0,0 +1,90 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="583",
+    message="More than one date the child should be placed for adoption has been reported, but an earlier decision has not been revoked. Check that both dates are required and add in the date the earlier decision was revoked. Note this may be in a previous year.",
```
SLornieCYC, line 8: Usually we've included the shorter error description as that matches what is shown to users in the DFE portal.

PR #720, opened 2023-11-07
Rule584

### LINKED ISSUE #618: Error code 584 [Added 2022 v1.1]

#### Error code
584
#### Description to display
Date of decision that the child should be placed for adoption this year is different from that recorded last year, but the decision to placed the child for adoption changed and the child should no longer be placed for adoption
If the date of the decision that the child should be placed for adoption in the current year is different to the date of the decision recorded in a previous year, then the decision to place the child for adoption must have changed, either in the current year or the previous year and the date should not be before the date of the earlier decision to place the child for adoption.
#### DfE suggested coding
If `<CURRENT_COLLECTION_YEAR>` -1 `<DATE_PLACED>` has been provided and `<DATE_PLACED_CEASED>` and or `<REASON_PLACED_CEASED>` is not NULL then `<CURRENT_COLLECTION_YEAR>` `<DATE_PLACED>` must be > `<CURRENT_COLLECTION_YEAR>` -1 `<DATE_PLACED>`
Apply for 2023 onwards.

WillLP-code:
closes #618 

Rule logic: fail rows where date_placed_current is earlier than date _placed_last in instances where there is no _reason_placed_ceased, and where there is a date_placed_last, also fails rows where current date_placed is empty, but date_placed_last is not, and date_placed_ceased and reason_placed_ceased are not


lac_validator/rules/lac2022_23/rule_584.py
```diff
@@ -0,0 +1,163 @@
...
+        # Rule logic: fail rows where date_placed_current is earlier than date _placed_last
+        # in instances where there is no _reason_placed_ceased, and where there is a date_placed_last
+
+        df = dfs["PlacedAdoption"]
+        df["index"] = df.index
+        pa_last = dfs["PlacedAdoption_last"]
+
+        df = df[
+            (df["DATE_PLACED_CEASED"].notna()) | (df["REASON_PLACED_CEASED"].notna())
+        ]
```
SLornieCYC: This filter is on the wrong dataframe.  We want to check **all** of the current year records against **the subset of the previous year** that were not ceased (date or reason).  Change `df` to `pa_last` and I think we're good for this part.

lac_validator/rules/lac2022_23/rule_584.py
```diff
@@ -0,0 +1,163 @@
...
+        df_merged = df.merge(
+            pa_last, how="left", on="CHILD", suffixes=("_current", "_last")
+        )
+        df_merged = df_merged[df_merged["DATE_PLACED_last"].notna()]
```
SLornieCYC: I think this is fine; but is there a benefit to dropping the nil-last-year rows by filtering after the merge like this, rather than making the merge `inner` instead of `left`?

lac_validator/rules/lac2022_23/rule_584.py
```diff
@@ -0,0 +1,163 @@
...
+        dp_earlier_list = list(
+            df_merged[
+                df_merged["DATE_PLACED_current"] < df_merged["DATE_PLACED_last"]
```
SLornieCYC: Think this needs to be `<=` rather than just `<` so we capture those in the current year that have the same start date as an ended decision in the previous year.

SLornieCYC (changes requested):

lac_validator/rules/lac2022_23/rule_584.py
```diff
@@ -0,0 +1,163 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="584",
+    message="If the date of the decision that the child should be placed for adoption in the current year is different to the date of the decision recorded in a previous year, then the decision to place the child for adoption must have changed, either in the current year or the previous year and the date should not be before the date of the earlier decision to place the child for adoption.",
```
SLornieCYC: Usually we've included the shorter error description as that matches what is shown to users in the DFE portal.

WillLP-code (commented):
@SLornieCYC I think that's everything you've asked for!

lac_validator/rules/lac2022_23/rule_584.py
```diff
@@ -0,0 +1,163 @@
...
+                "DATE_PLACED": "01/02/2020",
+                "DATE_PLACED_CEASED": "01/02/2020",
+                "REASON_PLACED_CEASED": "xx",
+            },  # 0 Fail, test neither null
+            {
+                "CHILD": "child2",
+                "DATE_PLACED": "01/02/2020",
+                "DATE_PLACED_CEASED": pd.NA,
+                "REASON_PLACED_CEASED": "xx",
+            },  # 1 Fail, test DPC null
```
SLornieCYC: I think the null `DPC` should be on the previous year record now for this test?

lac_validator/rules/lac2022_23/rule_584.py
```diff
@@ -0,0 +1,163 @@
...
+            {
+                "CHILD": "child3",
+                "DATE_PLACED": "01/02/2020",
+                "DATE_PLACED_CEASED": "01/02/2020",
+                "REASON_PLACED_CEASED": pd.NA,
+            },  # 2 Fail, test RPC null
```
SLornieCYC: I think the null `RPC` should be on the previous year record now for this test?

lac_validator/rules/lac2022_23/rule_584.py
```diff
@@ -0,0 +1,163 @@
...
+            {
+                "CHILD": "child4",
+                "DATE_PLACED": "01/02/2020",
+                "DATE_PLACED_CEASED": pd.NA,
+                "REASON_PLACED_CEASED": pd.NA,
+            },  # 3 Pass
```
SLornieCYC: This record fails now so comment is wrong.  Unless the intent is to test null for both `DEC` **and** `RPC` in which case those nulls should be on the previous year record now.

SLornieCYC (changes requested):

SLornieCYC (approved):
Think this looks right now.

PR #721, opened 2023-11-08
1016 working

### LINKED ISSUE #693: Error code 1016 [Added 2023 v1.2]
**Note: This rule was added by DFE in the 2023/24 v1.2 validation check document.**

#### Error code
1016
#### Description to display
Care leaver information is not required for 17- or 18-year olds who cease to be looked after, after their birthday
If the child ceases to be looked after in the year, but is still looked after on or after their 17th or on their 18th birthday, then information on their activity and accommodation is not required.
Please delete the OC3 information.
#### DfE suggested coding
If `<DOB>` < 19 and >= to 17 years prior to `<COLLECTION_END_DATE>`
and
current episode `<DEC>` is present and after the birthdate then
`<IN_TOUCH>`, `<ACTIV>` and `<ACCOM>` should not be provided

WillLP-code:
closes #693 

Adds 1016 to 23/24
Uses np.timedeltas to calculate ages, which I think is fine, but I don't know if there's some obscure reason where it may not catch edge cases say if someone has been alive for 19 years but somehow isn't 19, or something, I'm not sure if there was a standard way of doing it for the validator I'm not aware of.

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_1016.py
+++ b/lac_validator/rules/lac2023_24/rule_1016.py
@@ -38,25 +38,28 @@ def validate(dfs):
 
         # Fails rows where the child is between 19 and 17 at the end of a return,
         # and their DEC is after their 17th birthday
-        df_merged['age at dec'] = (df_merged['DEC'] - df_merged['DOB']) / np.timedelta64(1, "Y")
-    
-        condition_17 = df_merged["DEC"] >= (df_merged["DOB"] + pd.offsets.DateOffset(years=17))
-        condition_18 = df_merged["DEC"] >= (df_merged["DOB"] + pd.offsets.DateOffset(years=18))
-        print(df_merged)
-        print(df_merged[condition_18])
+        df_merged["age at dec"] = (
+            df_merged["DEC"] - df_merged["DOB"]
+        ) / np.timedelta64(1, "Y")
+
+        condition_17 = df_merged["DEC"] >= (
+            df_merged["DOB"] + pd.offsets.DateOffset(years=17)
+        )
+        condition_18 = df_merged["DEC"] >= (
+            df_merged["DOB"] + pd.offsets.DateOffset(years=18)
+        )
 
         df_errors = df_merged[
             (
-                (df_merged["AGE_AT_CE"] >= np.timedelta64(17, "Y")) & 
-                 (df_merged["AGE_AT_CE"] < np.timedelta64(18, "Y"))  & 
-                 condition_17
-                 )
+                (df_merged["AGE_AT_CE"] >= np.timedelta64(17, "Y"))
+                & (df_merged["AGE_AT_CE"] < np.timedelta64(18, "Y"))
+                & condition_17
+            )
             | (
-                (df_merged["AGE_AT_CE"] >= np.timedelta64(18, "Y")) & 
-                 (df_merged["AGE_AT_CE"] < np.timedelta64(19, "Y")) & 
-                 condition_18
-                 )
-            
+                (df_merged["AGE_AT_CE"] >= np.timedelta64(18, "Y"))
+                & (df_merged["AGE_AT_CE"] < np.timedelta64(19, "Y"))
+                & condition_18
+            )
         ]
 
         df_errors = df_errors[
@@ -81,14 +84,14 @@ def test_validate():
                 "ACCOM": "Y",
             },
             {
-                "CHILD": "child2", 
+                "CHILD": "child2",
                 "DOB": "01/01/2001",
                 "IN_TOUCH": pd.NA,
                 "ACTIV": pd.NA,
                 "ACCOM": pd.NA,
             },
             {
-                "CHILD": "child3", 
+                "CHILD": "child3",
                 "DOB": "01/01/2000",
                 "IN_TOUCH": "Y",
                 "ACTIV": "Y",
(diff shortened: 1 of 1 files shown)


SLornieCYC:
Can't see any problem with using np.timedeltas.  I think most of the previous implementations around age have used pd.offsets.DateOffset - for example rule 1007 which is very similar to this one. (No I don't know why DFE have written the validation rules this way; I'm sure it makes sense to them!).

https://github.com/data-to-insight/csc-validator-be-903/blob/77d4908cd90792ecb77dea9884c6ef49eefa032b/lac_validator/rules/lac2022_23/rule_1007.py#L33-L38

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -0,0 +1,85 @@
...
+                "DOB": "01/01/2002",
+                "IN_TOUCH": "Y",
+                "ACTIV": "Y",
+                "ACCOM": "Y",
+            },
+        ]
+    )
+
+    fake_data["DOB"] = pd.to_datetime(fake_data["DOB"], format="%d/%m/%Y")
+    metadata = {"collection_end": pd.to_datetime("01/01/2020", format="%d/%m/%Y")}
```
SLornieCYC: Doesn't this need to be in the main `validate()` function? I don't think the dataframes from ingress are converted due to some weird bit in the DFE specification for partial/unknown dates.

WillLP-code: I actually didn't know ingress didn't convert to datetimes, I think there's a bunch of times I'll need to fix this then!

SLornieCYC: See this comment on a previous issue:  https://github.com/data-to-insight/csc-validator-be-903/issues/654#issuecomment-1679033062.  Although I have no clue how (or if) DFE validate such data through the rest of the date-based rules (I think we've just accepted we're ignoring partial dates in this tool!).

SLornieCYC (changes requested):
This needs to include a check against the Episodes file to determine whether the child left care after their in-year birthday.  (We only want to flag those who have OC3 data **and** their last Episodes `DEC` was **after** their 17th or 18th birthday).

WillLP-code (commented):
@SLornieCYC I've had a look at this but I might need to look again tomorrow. I don't know how I missed the DEC part of the rule in my original attempt!

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -0,0 +1,134 @@
...
+    affected_fields=["IN_TOUCH", "ACTIV", "ACCOM"],
+)
+def validate(dfs):
+    # If <DOB> < 19 and >= to 17 years prior to <COLLECTION_END_DATE>
+    # and
+    # current episode <DEC> is present and after the birthdate then
+    # <IN_TOUCH>, <ACTIV> and <ACCOM> should not be provided
+    if ("OC3" not in dfs) & ("Episodes" not in dfs):
+        return {}
+    else:
```
SLornieCYC: This needs to be `or` rather than `&`.

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -0,0 +1,134 @@
...
+
+        episodes["DEC"] = pd.to_datetime(
+            episodes["DEC"], format="%d/%m/%Y", errors="coerce"
+        )
+
+        OC3 = OC3.reset_index()
+
+        OC3["AGE_AT_CE"] = collection_end - OC3["DOB"]
+
+        df_merged = OC3.merge(episodes, on="CHILD", how="left")
```
SLornieCYC: _Think_ this can be an inner join because we only care about OC3 records who have episodes in current year.  (They could be 17-18yrs and have left care in previous year and therefore have no episodes this year).

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Fails rows where the child is between 19 and 17 at the end of a return,
+        # and their DEC is after their 17th birthday
+        df_errors = df_merged[
+            (df_merged["AGE_AT_CE"] < np.timedelta64(19, "Y"))
+            & (df_merged["AGE_AT_CE"] >= np.timedelta64(17, "Y"))
+            & (df_merged["DEC"] >= (df_merged["DOB"] + np.timedelta64(17, "Y")))
```
SLornieCYC: The last filter here needs to accommodate 17th **_or 18th_** birthday as it could be either of these.

SLornieCYC (changes requested):

lac_validator/rules/lac2023_24/rule_1016.py
WillLP-code: So, I could be wrong, but what I've done is take anyone who is between 17 and 19 and who has a DEC after their 17th birthday, so that will include children who have had their 17th or 18th birthdays. Wouldn't adding a check for those who have a DEC after their 18th birthday be redundant?

SLornieCYC: The key thing here is that they need to have left care on or after **the in-year birthday** to flag on this rule.

For example:

- Child turned 18 on 1 September
- Child left care on 31 August (before birthday)

At the moment this case will return **true**, **true**, **true** for the filter (fail) because the last part checks the 17th birthday only, whereas we actually want the third part to be checking their 18th birthday because they are turning 18 during the year (and which should return false).

[`AGE_AT_CE` < 19, `AGE_AT_CE` >= 17, **`DEC` >= 17th birthday**] 

Here are some other examples that I think should pass (any false) or fail (all criteria true):

### Pass
- Child turned 17 on 1 September
- Child left care on 31 August (before birthday)

### Fail
- Child turned 17 on 1 September
- Child left care on 2 September (after in-year birthday)

### Fail
- Child turned 18 on 1 September
- Child left care on 2 September (after in-year birthday)

WillLP-code: If I am right in my interpretation of the above, the latest commit should pass and fail according to the above.

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -107,21 +110,45 @@ def test_validate():
...
     fake_epi = pd.DataFrame(
         [
-            {"CHILD": "child1", "DEC": "01/01/2020"},
+            {"CHILD": "child1", "DEC": "31/03/2020"},
             {"CHILD": "child2", "DEC": pd.NA},
             {"CHILD": "child3", "DEC": pd.NA},
             {"CHILD": "child4", "DEC": "01/01/2020"},
             {"CHILD": "child6", "DEC": pd.NA},
             {"CHILD": "child4", "DEC": pd.NA},
+            {"CHILD": "child8", "DEC": "31/08/20020"},
```
SLornieCYC: Typo in year; think this should be 2020?

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -107,21 +110,45 @@ def test_validate():
...
+            {"CHILD": "child9", "DEC": "02/09/2020"},
+            {"CHILD": "child10", "DEC": "02/09/2020"},
```
SLornieCYC: On second thought I think the dates used for these three tests make this a bit harder to comprehend because you've set `collection_end` = 31/03/2020 and then using `DEC` after that.  I'd suggest making the `DEC` between 01/04/2019 and 31/03/2020 (so within the same return year as `collection_end`) and adjust the DOBs accordingly.

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -38,10 +38,13 @@ def validate(dfs):
...
         # and their DEC is after their 17th birthday
+        condition_17 = df_merged["DEC"] >= (df_merged["DOB"] + np.timedelta64(17, "Y"))
+        condition_18 = df_merged["DEC"] >= (df_merged["DOB"] + np.timedelta64(18, "Y"))
+
         df_errors = df_merged[
             (df_merged["AGE_AT_CE"] < np.timedelta64(19, "Y"))
             & (df_merged["AGE_AT_CE"] >= np.timedelta64(17, "Y"))
-            & (df_merged["DEC"] >= (df_merged["DOB"] + np.timedelta64(17, "Y")))
+            & (condition_17 | condition_18)
         ]
```
SLornieCYC: This looks better to me but I think it needs to be structured like this to avoid `condition_17` applying when the child turns 18 duing the year.  But with however many extra brackets python needs for that expression.

`AGE_AT_CE` >= 17 & `AGE_AT_CE` < 18 & `condition_17` |  `AGE_AT_CE` >= 18 & `AGE_AT_CE` < 19 & `condition_18`

lac_validator/rules/lac2023_24/rule_1016.py
```diff
@@ -131,4 +158,4 @@ def test_validate():
 
     result = validate(fake_dfs)
 
-    assert result == {"OC3": [0, 3]}
+    assert result == {"OC3": [0, 3, 8, 9]}
```
SLornieCYC: As covered in above comment about `condition_17` I think some of these may be false-positives due to turning 18 but the logic is currently seeing `DEC` > 17.

SLornieCYC (changes requested):

lac_validator/rules/lac2023_24/rule_1016.py
WillLP-code: So, I've made explicit in the comments where I think rows should pass and fail based on my understanding of the rule. I've also had to change the condition 17/18 to use pandas date offsets because for some periods being having a dec the same date 18 years after a DOB didn't count as the right number of years because no.timedelta has years as exactly 365 86,400s days, so for some periods leap seconds messed up calculations (namely child 4 with a DOB of 01/01/2002 and a dec of 01/01/2020 was passing when they should fail as they're 18 ON their DEC)

SLornieCYC: Thanks Will; I think this one looks right to me now too.  I've pushed some amendments to the tests.

SLornieCYC (approved):

PR #723, opened 2023-11-08
174 updated gender code 1 to F

### LINKED ISSUE #696: Error code 174 [Amended 2024 v1.0]
#### Error code
174
#### Description to display
Mother’s child date of birth is recorded but sex shows that the child is a male
Date of birth of child, can only be entered if sex is female and mother status = 1.
#### DfE suggested coding
If `<SEX>` = ~~'1'~~ **‘F’** then `<MC_DOB>` should not be provided

WillLP-code:
closes #696
Potential to add a check for "1" as well as "F". I'm assuming the DFE changed gender codes?

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_174.py
+++ b/lac_validator/rules/lac2023_24/rule_174.py
@@ -1,0 +1,41 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="174",
+    message="Mother's child date of birth is recorded but gender shows that the child is a male.",
+    affected_fields=["SEX", "MC_DOB"],
+)
+def validate(dfs):
+    if "Header" not in dfs:
+        return {}
+    else:
+        header = dfs["Header"]
+
+        child_is_male = header["SEX"] == "F"
+        mc_dob_recorded = header["MC_DOB"].notna()
+
+        error_mask = child_is_male & mc_dob_recorded
+
+        validation_error_locations = header.index[error_mask]
+
+        return {"Header": validation_error_locations.to_list()}
+
+
+def test_validate():
+    import pandas as pd
+
+    fake_data = pd.DataFrame(
+        {
+            "SEX": ["F", "F", "2", "2", 1, 1],
+            "MC_DOB": [pd.NA, "19/02/2010", pd.NA, "19/02/2010", pd.NA, "19/02/2010"],
+        }
+    )
+
+    fake_dfs = {"Header": fake_data}
+
+    result = validate(fake_dfs)
+
+    assert result == {"Header": [1]}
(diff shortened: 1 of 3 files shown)


lac_validator/rules/lac2023_24/rule_174.py
```diff
@@ -0,0 +1,41 @@
...
+
+        return {"Header": validation_error_locations.to_list()}
+
+
+def test_validate():
+    import pandas as pd
+
+    fake_data = pd.DataFrame(
+        {
+            "SEX": ["F", "F", "2", "2", 1, 1],
```
SLornieCYC: I think you've got your "F"s in the wrong place here?  Previously 1 = M and 2 = F.  Probably best to directly convert all of the 1s and 2s from original rule to the corresponding letter:

https://github.com/data-to-insight/csc-validator-be-903/blob/77d4908cd90792ecb77dea9884c6ef49eefa032b/lac_validator/rules/lac2022_23/rule_174.py#L32-L33

SLornieCYC:
@WillLP-code Yes DFE have created a new codeset that clarifies what they're collecting (previously sex/gender terms were used interchangeably within and between returns).  Note that this comes in from the **2024/2025** return (not 2023/2024) so you need to move your file forward a year in the folder structure!

lac_validator/rules/lac2023_24/rule_174.py
WillLP-code: Got confused by this 'If <SEX> = ~~'1'~~ ‘F’ then <MC_DOB> should not be provided' which made me think it was just swapping current 1s for new Fs!

SLornieCYC: For a moment I thought I'd written it wrong when I added the issue. But no, DFE have it wrong in their new guidance! 

### Previous rule and gender coding

image

image


### New rule and gender coding

image

image

WillLP-code: wow! Well I'm glad I coded it right, even if it's wrong!

lac_validator/rules/lac2024_25/__init__.py
```diff
@@ -0,0 +1,21 @@
+from pathlib import Path
+
+from lac_validator.rule_engine import RuleDefinition, YearConfig
+from lac_validator.rules.lac2023_24 import registry as prev_registry
```
SLornieCYC, line 4: Is this right? All I did was copy this file from the previous year folder and update this to refer to the new previous year.  Will it still cascade back through the previous two years and pick up all the unchanged rules from the lac2022_23 folder?

SLornieCYC: Meant to tag you in this one @WillLP-code!

WillLP-code: So, the 23/24 rules build from the _init_ file, which, you are right, goes through the old rules and adds them to all the ones from the current years folder!

SLornieCYC (approved):

PR #724, opened 2023-11-08
218 updated 23/24 removed H5, added test

### LINKED ISSUE #697: Error code 218 [Amended 2024 v1.0]
#### Error code
218
#### Description to display
Ofsted Unique reference number (URN) is required.
Please ensure Ofsted URN is recorded for ongoing placements at 1 April of the current collection year and also for new episodes from 1 April of the current collection year where the reason for new episode is P, T, B, U or S.
Ofsted URN should not be provided for the placement types listed in rule 222, for confidential placements, placements outside of England.
#### DfE suggested coding
Where `<PL>` not in ~~'H5';~~ ‘P1’; ~~'P2';~~ ‘P3’; ‘R1’; ‘R2’; ‘R5’; ‘T0’; ‘T1’; ‘T2’; ‘T3’; ‘T4’ ‘Z1’
And `<PL_LA>` not in ‘CON’; ‘NIR’ ‘SCO’; ‘WAL’; ‘NUK’
And `<DEC>` > 31/03/2015 or Null then `<URN>` should be provided

WillLP-code:
closes #697 

Removed H5 from list of allowed URN codes, adds a test for H5 to fail.


lac_validator/rules/lac2023_24/rule_218.py
```diff
@@ -0,0 +1,158 @@
...
+
+    fake_data_eps = pd.DataFrame(
+        [
+            {
+                "CHILD": "0000",
+                "PLACE": "H5",
+                "DEC": pd.NA,
+                "PL_LA": "o",
+                "URN": pd.NA,
+            },  # 4 fail
```
SLornieCYC: It's not so much that we want H5 to fail, but that H5 (and P2) is no longer a valid code in the return so should never even appear in the data (which is handled by another validation rule).  Simpler to just keep the original coded test here but substitute H5 for another item in `pl_list`?

Also this is test index 0 not 4! ;)

SLornieCYC:
Note that this change comes in from the **2024/2025** return (not 2023/2024) so you need to move your file forward a year in the folder structure!

lac_validator/rules/lac2023_24/rule_218.py
SLornieCYC, line 1: This one is fine now, apart from the change comes in from the 2024/2025 return (not 2023/2024) so the rule file needs moving forward a year in the folder structure.

SLornieCYC (changes requested):

SLornieCYC (approved):

PR #725, opened 2023-11-08
222 ammended for 23/24 return with test

### LINKED ISSUE #698: Error code 222 [Amended 2024 v1.0]
#### Error code
222
#### Description to display
Ofsted Unique reference number (URN) should not be recorded for this placement type.
Ofsted URN should not be recorded for the following placement types:
P1 - Placed with own parent(s) or other person with parental responsibility
~~P2 - Independent living, for example, in flat, lodgings, bedsit, bed and breakfast (B & B) or with friends, with or without formal support~~
P3 - Residential employment
R1 - Residential care home
R2 – National Health Service (NHS) / health trust or other establishment providing medical or nursing care
R5 - Young offender institution or prison
T0 - All types of temporary move
T1 - Temporary periods in hospital
T2 -Temporary absences of the child on holiday
T3 - Temporary accommodation whilst normal foster carer(s) is/are on holiday
T4 - Temporary accommodation of seven days or less, for any reason, not covered by codes T1 to T3
Z1 - Other placements.
#### DfE suggested coding
If `<URN>` provided and `<URN>` not = ‘XXXXXX’, and where `<PL>` = ‘P1’; ~~'P2';~~ ‘P3’; ‘R1’; ‘R2’; ‘R5’; ‘T0’ ‘T1’; ‘T2’; ‘T3’; ‘T4’ or Z1 then <URN> should not be provided

WillLP-code:
closes #698


lac_validator/rules/lac2023_24/rule_222.py
```diff
@@ -47,6 +48,7 @@ def test_validate():
             {"PLACE": "T1", "URN": pd.NA},  # 3
             {"PLACE": "R1", "URN": "Whatever"},  # 4  Fail
             {"PLACE": "T2", "URN": "Whatever"},  # 5  Fail
+            {"PLACE": "P2", "URN": "Whatever"},  # 6  Pass
```
SLornieCYC: This line isn't needed because this rule change is due to P2 ceasing to be a valid code **at all** in 2024/25; all we're doing is updating the rule to remove the reference to a now redundant item.  (Checking whether used `PLACE` codes are valid is handled by another rule).

SLornieCYC (changes requested):
This change is fine apart from one redundant line - and the fact that it needs to be added in the **2024/25** year!  (Rather than amended in 2023/24 when P2 is still a valid code)

lac_validator/rules/lac2023_24/rule_222.py
WillLP-code: I'll do this when the 218 change is merged to main so I can merge main back into this branch, and avoid making the 24/25 folder and init in multiple branches and risking conflicts!

SLornieCYC (approved):
Think this is good now!

Issue #634, opened 2023-05-02
Error code SW13STG1 [Added 2023 v1.1]


SLornieCYC:
**Note: This rule was added by DFE in the 2023/24 v1.1 validation check document. It is dependent on #637.**

#### Error code
SW13STG1
#### Description to display
The reason for social worker change is either not valid or has not been entered.
#### DfE Long-form description of error
The reason for social worker change field must not be blank, and code must be taken from the list supplied in the DfE children looked after collection guide.
Valid codes are MANAGE, FCONTA, LEFTRL, ORGRST, TSPROC, ABSENC, CHCHAN, PCCHAN, SWDIED, OTHERS.
Codes must all be upper case.
#### DfE suggested coding
`<SW_REASON>` must be provided and be a valid value
For allowed values please see associated code set.

For further info, refer to the DfE validation document [here](https://www.gov.uk/government/publications/children-looked-after-return-2023-to-2024-technical-specifications)

PR #727, opened 2023-11-14
Feature/sw ingress

### LINKED ISSUE #637: New data table needed from 2023/24 - Social Worker Episodes
Starting from the 2023/24 collection the SSDA903 return includes a new module for "Social Worker Episodes".  To include this in the validator tool will need to amend config.py and ingress.py as well as make associated changes to the tool front-end (file upload etc).  See DFE return technical specification for further details: [here](https://www.gov.uk/government/publications/children-looked-after-return-2023-to-2024-technical-specifications)

WillLP-code:
closes #637 #634 

Adds a CLI command to run the tool on the fake data in the tests folder, used to make sure the new SW_Episodes file can be read by the BE. 

Adds the sw_episodes form to the config file.

Adds a minimal fake version of the sw_episodes file to the test/fake_data directory.

Adds SW13STG1 as a means of testing ingress and functionality of adding SW List to ingress


lac_validator/__main__.py
```diff
@@ -137,6 +137,76 @@ def run_all(p4a_path, ad1_path, ruleset, select):
...
+        selected_rules=None,
+    )
+
+    click.echo(v.dfs)
+
+    results = v.ds_results
+
+    r = Report(results, ruleset_registry)
+    full_issue_df = lac_validator.create_issue_df(r.report, r.error_report)
+    click.echo(full_issue_df[full_issue_df["tables_affected"] == "SW_Episodes"])
```
SLornieCYC: Should this filter be hard-coded here or ought it be defined as an argument or option in the CLI function?

lac_validator/config.py
```diff
@@ -59,4 +59,5 @@
     # Note: for PrevPerm, LA_PERM will usually be blank and shouldn't be used directly
     "PrevPerm": ["CHILD", "DOB", "PREV_PERM", "LA_PERM", "DATE_PERM"],
     "Missing": ["CHILD", "DOB", "MISSING", "MIS_START", "MIS_END"],
+    "SW_Episodes": ["CHILD", "DOB", "SW_ID", "SW_DECOM", "SW_DEC", "SW_REASON"],
```
SLornieCYC: All of the other file names in the tool are done without _ for spaces (e.g. `PlacedAdoption` rather than `Placed_Adoption`) so think should be `SWEpisodes` here (and throughout tool) for consistency.

lac_validator/__main__.py
WillLP-code: Oh I didn't even think about this, I just play around with that particular CLI command as and when to run whatever I need offline and see what happens. I'll make it a bit more universal!

lac_validator/config.py
WillLP-code: That's a great shout! I wonder if we need to think about having consistency between the CIN and 903, as I used the CIN structure here, but I think that job is bigger than it is useful. I'll fix this now!

SLornieCYC (approved):
Looks good to me!

PR #736, opened 2023-12-13
SW16bSTG2

### LINKED ISSUE #708: Error code SW16bSTG2 [Added 2024 v1.0]
**Note: This rule was added by DFE in the 2024/25 v1.0 validation check document. It is dependent on #637.**

#### Error code
SW16bSTG2
#### Description to display
The social worker episode start date does not match the open episode at the end of last year.
There was an open social worker episode at the end of last year, the first social worker episode this year must be consistent.
Check the social worker episode start date reported in both years.
#### DfE suggested coding
If present, where there is a previous collection year’s final social worker episode, and `<SW_DEC>` not provided, then first social worker episode of this year’s collection `<SW_DECOM>` be the same as the previous episode

WillLP-code:
closes #708 

Pretty much the same logic as SW16cSTG2

Orders previous SWE decom and keeps only the last, orders current SWE decom and keeps only the first, then if there is no DEC in previous SWE, checks they have the same SWE DECOM, failing if they do not.

### DIFF
--- a/lac_validator/rules/lac2024_25/rule_SW16bSTG2.py
+++ b/lac_validator/rules/lac2024_25/rule_SW16bSTG2.py
@@ -1,0 +1,127 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="SW16bSTG2",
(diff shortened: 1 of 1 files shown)


SLornieCYC (approved):
I think this one also looks fine.

PR #738, opened 2023-12-13
SW15STG2

### LINKED ISSUE #706: Error code SW15STG2 [Added 2024 v1.0]
**Note: This rule was added by DFE in the 2024/25 v1.0 validation check document. It is dependent on #637.**

#### Error code
SW15STG2
#### Description to display
Reason for social worker change is missing.
Reason for social worker change can only be blank if it is the first social worker episode for the reporting period and the reason was not collected.
In all other instances it should be provided.
#### DfE suggested coding
If `<SW_REASON>` is NULL, then `<DECOM>` must be < `<COLLECTION_START_DATE>` minus one year

WillLP-code:
closes  #706

Coded as per the DfE suggested coding, but I'm not convinced the suggested coding means the same thing as the long form description of the error

> DfE Long-form description of error
> Reason for social worker change can only be blank if it is the first social worker episode for the reporting period and the reason was not collected.
> In all other instances it should be provided.
> 
> DfE suggested coding
> If <SW_REASON> is NULL, then <DECOM> must be < <COLLECTION_START_DATE> minus one year

### DIFF
--- a/lac_validator/rules/lac2024_25/rule_SW15STG2.py
+++ b/lac_validator/rules/lac2024_25/rule_SW15STG2.py
@@ -1,0 +1,62 @@
+import pandas as pd
+import numpy as np
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="SW15STG2",
+    message="Reason for social worker change is missing.",
+    affected_fields=["SW_REASON"],
+)
+def validate(dfs):
+    if "SWEpisodes" not in dfs:
+        return {}
+    else:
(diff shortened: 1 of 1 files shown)


SLornieCYC:
The coding description and the long form description are compatible because from the 2024/25 return the SWEpisodes module covers **two full calendar years** of episode data (vs one year for other modules, and SWEpisodes for its first collection in 2023/24).

This might also be a partial reflection that because it's a new requirement to collect this data, LAs may not have it for older records.

SLornieCYC (approved):
This one looks fine too!

Issue #705, opened 2023-10-10
Error code SW14STG2 [Added 2023 v1.2]


SLornieCYC:
**Note: This rule was added by DFE in the 2023/24 v1.2 validation check document. It is dependent on #637.**

#### Error code
SW14STG2
#### Description to display
Social worker episode date ceased is blank.
#### DfE Long-form description of error
Only the last social worker episode in the reporting period can be blank.
This can only be when the episode is ongoing at the 31 March.
#### DfE suggested coding
If `<SW_DEC>` is null, then it must belong to the latest social worker episode.

For further info, refer to the DfE validation document [here](https://www.gov.uk/government/publications/children-looked-after-return-2023-to-2024-technical-specifications)

SLornieCYC:
Note the coding description above has been taken from the 2024/25 v1.0 document because that column was missing from the 2023/24 v1.2 document.  (But the rule is not marked as amended in 2024/25 v1.0 so should be consistent).

PR #741, opened 2023-12-14
SWo3STG1

### LINKED ISSUE #671: Error code SW03STG1 [Added 2023 v1.1]
**Note: This rule was added by DFE in the 2023/24 v1.1 validation check document. It is dependent on #637.** (Rule amended in the 2023/24 v1.2 validation check document).

#### Error code
SW03STG1
#### Description to display
For each social worker episode, information should be complete
If any social worker episode data items are provided then all should be provided, however date episode ceased may be null if the episode is ongoing at 31 March.
#### DfE suggested coding
If any of the social worker episode data items are present, the following data items must be present i.e. `<SW_ID>`, `<SW_DECOM>`, `<SW_DEC>`, `<SW_REASON>`

WillLP-code:
Closes #671

Fails children who have one of any of SW ID, SW DECOM, SW DEC, or SW REASON missing BUT allows children to pass if they only have SW DEC missing if they were looked after at 31 march (checked with the episode file having a null DEC and LA not being V3 or V4

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_SW03STG1.py
+++ b/lac_validator/rules/lac2023_24/rule_SW03STG1.py
@@ -1,0 +1,125 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="SW03STG1",
+    message="For each social worker episode, information should be complete.",
+    affected_fields=["SW_ID", "SW_DECOM", "SW_DEC", "SW_REASON"],
+)
+def validate(dfs):
+    if ("SWEpisodes" not in dfs) | ("Episodes" not in dfs):
+        return {}
+    else:
+        SWE = dfs["SWEpisodes"]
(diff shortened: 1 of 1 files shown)


SLornieCYC:
This is one of those where I think you've overcomplicated the rule coding @WillLP-code.  For the return LAs should only be providing `SWEpisodes` for children in care at 31 March anyway, so there isn't really a need to merge to `Episodes` and check that for this rule.  (Although right now I can't actually see any rule that flags data submitted on children not in care at 31 March - but that's a DFE problem not ours!).

I think I'd be tempted to deviate slightly from the DFE coding and test there is content in all of the `SWEpisode` fields apart from `DEC` (which may be null).  There are other rules that specifically test whether `DEC` is allowed to be blank for any particular episode (#747, #750, #705).

SLornieCYC (approved):
Looks good to me!

PR #742, opened 2023-12-15
SW04STG1

### LINKED ISSUE #672: Error code SW04STG1 [Added 2023 v1.1]

#### Error code
SW04STG1
#### Description to display
Date social worker episode began is not a valid date.
If completed the date the social worker episode began must be a valid date and should be entered in the format DD/MM/YYYY.
Acceptable values where date is not known will include zz, where zz is day or month value or zzzz is year value.
Do not leave this item blank.
#### DfE suggested coding
If provided `<SW_DECOM>` must be in a valid format and contain a valid date
Format should be DD/MM/YYYY, or one or more elements of the date can be replaced by zz if part of the date element is not known.

WillLP-code:
closes #672 

This is another slightly weird one. I cannibalised the code from 632 to do it, but have changed it to allow any and all date values to be ZZs in line with the wording of the rule, but I'm not sure if that's how the guidance is meant to be interpreted.

I do wonder if this is possible without a for loop. I tried to do it from scratch but realised if 632 was in, it may be the best way of doing it.

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_SW04STG1.py
+++ b/lac_validator/rules/lac2023_24/rule_SW04STG1.py
@@ -1,0 +1,76 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="SW04STG1",
+    message="Date social worker episode began is not a valid date..",
+    affected_fields=["SW_DECOM"],
+)
+def validate(dfs):
+    if "SWEpisodes" not in dfs:
+        return {}
+    else:
+        df = dfs["SWEpisodes"]
+
+        # function to check that date is of the right format
+        def valid_date(dte):
+            try:
+                lst = dte.split("/")
+            except AttributeError:
+                return pd.NaT
+            # Preceding block checks for the scenario where the value passed in is nan/naT
+
+            # date should have three elements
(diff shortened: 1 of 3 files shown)


SLornieCYC:
Would it make sense now to move the `valid_date` function out of the individual rules and share that as common code between all of them?  Rather than having separate and slightly different instances directly coded in each rule?

lac_validator/rules/lac2023_24/rule_SW04STG1.py
```diff
@@ -0,0 +1,41 @@
...
+        {
+            "SW_DECOM": ["ZZ/ZZ/ZZZZ", "01/01/2001", "zz", "01/01/ZZZZ", pd.NA],
+        }
+    )
+
+    fake_dfs = {"SWEpisodes": fake_data}
+
+    result = validate(fake_dfs)
+
+    assert result == {"SWEpisodes": [2, 4]}
```
SLornieCYC: I wouldn't expect row 4 to fail here because the rule starts by saying "If provided...".  A null `SW_DECOM` is perfectly fine here and handled in another rule.

SLornieCYC (approved):
Made a small tweak; think this is good now.

PR #750, opened 2023-12-19
SW11STG2

### LINKED ISSUE #679: Error code SW11STG2 [Added 2023 v1.1]

#### Error code
SW11STG2
#### Description to display
Social worker episode has ended before 31 March and no subsequent social worker episode has been provided.
The social worker episode has ended before 31 March.
All social worker episodes should be provided for the time period 1 April to 31 March for a child looked after on 31 March, including periods when the child was no longer looked after if they still had a social worker,
Missing social worker episodes should be added.
#### DfE suggested coding
If `<SW_DEC>` < `<COLLECTION_END_DATE>`
then next social worker episode `<SW_DECOM>` must be present and must be = `<SW_DEC>`

WillLP-code:
closes #679

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_SW11STG2.py
+++ b/lac_validator/rules/lac2023_24/rule_SW11STG2.py
@@ -6,7 +6,7 @@ from lac_validator.rule_engine import rule_definition
 @rule_definition(
     code="SW11STG2",
     message="Social worker episode has ended before 31 March and no subsequent social worker episode has been provided.",
-    affected_fields=["SW_DECOM"],
+    affected_fields=["SW_DEC"],
 )
 def validate(dfs):
     if "SWEpisodes" not in dfs:
@@ -32,6 +32,7 @@ def validate(dfs):
             df_lead,
             left_on="index",
             right_on="level_0",
+            how="outer",
             suffixes=("", "_prev"),
         )
 
--- a/lac_validator/rules/lac2023_24/rule_SW11STG2.py
+++ b/lac_validator/rules/lac2023_24/rule_SW11STG2.py
@@ -1,0 +1,70 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="SW11STG2",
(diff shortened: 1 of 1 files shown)


lac_validator/rules/lac2023_24/rule_SW11STG2.py
```diff
@@ -0,0 +1,70 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="SW11STG2",
+    message="Social worker episode has ended before 31 March and no subsequent social worker episode has been provided.",
+    affected_fields=["SW_DECOM"],
```
SLornieCYC: I think this should be `SW_DEC` because it's the _absence_ of a following `SW_DECOM` that's the problem (and we can't highlight a missing data item!).

lac_validator/rules/lac2023_24/rule_SW11STG2.py
```diff
@@ -0,0 +1,70 @@
...
+            suffixes=("", "_prev"),
+        )
+
+        before_end = m_df[m_df["SW_DEC_prev"] < collection_end]
+
+        error_rows = before_end[
+            before_end["SW_DECOM"].isna()
+            | (before_end["SW_DECOM"] != before_end["SW_DEC_prev"])
+        ]["index"]
+
```
SLornieCYC: Took me a little while to get my head around this. So you're:

1) Merging all episodes to a 1-row-shifted copy of the same, to get the data from the previous episode.
2) Only capturing rows from the mere where the previous episode has ended (before the collection end).
3) Checking that the rows have either no `SW_DECOM`, or the `SW_DECOM` for the row is different to the previous episode `SW_DEC`.

For the first part of 3) (no `SW_DECOM` after a previous episode) does this merge need to be specified as outer, so that it actually captures rows that do not have a subsequent episode row?

SLornieCYC (changes requested):

SLornieCYC (approved):
This one looks fine! (Sorry it took so long for me to come back and re-review it).

PR #751, opened 2023-12-19
SW12STG2

### LINKED ISSUE #680: Error code SW12STG2 [Added 2023 v1.1]

#### Error code
SW12STG2
#### Description to display
If a child is looked after on 31 March then all social worker episodes for the full previous year should be reported, even those whilst they were not looked after
The child has had more than one period of care in the year. There is no social worker episode reported for the time prior to the latest period of care.
Social worker episode data should be provided for the full period 1 April to 31 March for any child looked after on 31 March. Note: This includes any periods when the child was not looked after but still had a social worker.
#### DfE suggested coding
For the child, if any previous non-respite `<PERIOD_OF_CARE>` exists where the last `<DEC>` is in the `<CURRENT_COLLECTION_YEAR>`
then for each period of care there must be at least one social worker episode where
`<SW_DECOM>` < last `<DEC>` in that `<PERIOD_OF_CARE>`

WillLP-code:
Closes #680 

I'm going to have to come back to this when I've done the other rules, it doesn't work and I need to think about how to calculate periods of non-respite care, separate them, and then calculate if SW episodes exist for each of them


SLornieCYC:
The `add_col_to_episodes_CONTINUOUSLY_LOOKED_AFTER` function in lac_validator/utils.py may be of help with determining non-respite periods of care (the code logic within, not the actual function).  E.g. using `RNE` = "S" and `REC` <> "X1" to identify breaks in periods of care.

I'm thinking something like this:

1) Make sure there is a SW episode where `SW_DECOM` <= episode `DEC` where `REC` <> "X1"
2) Make sure there is a SW episode where `SW_DECOM` >= episode `DECOM` where `RNE` = "S"
3) Exclude episodes from the above where `LS` = V3/V4

I'm assuming another rule somewhere is testing "has a SW episode if open at end of year".

WillLP-code:
> The `add_col_to_episodes_CONTINUOUSLY_LOOKED_AFTER` function in lac_validator/utils.py may be of help with determining non-respite periods of care (the code logic within, not the actual function). E.g. using `RNE` = "S" and `REC` <> "X1" to identify breaks in periods of care.
> 
> I'm thinking something like this:
> 
> 1. Make sure there is a SW episode where `SW_DECOM` <= episode `DEC` where `REC` <> "X1"
> 2. Make sure there is a SW episode where `SW_DECOM` >= episode `DECOM` where `RNE` = "S"
> 3. Exclude episodes from the above where `LS` = V3/V4
> 
> I'm assuming another rule somewhere is testing "has a SW episode if open at end of year".

@SLornieCYC I've made a further attempt at this, trying to take into account what you said. I haven't written a test yet because I wanted to check if this was along the lines of what you had in mind? I'm happy for you to change/completely re-do this if you want to and find you have time

lac_validator/rules/lac2023_24/rule_SW12STG2.py
```diff
@@ -0,0 +1,79 @@
...
+        df["DEC_IN_YEAR"] = (df["DEC"] >= collection_start) & (
+            df["DEC"] <= collection_end
+        )
+
+        df = pd.merge(epi, SWE, how="left", on="CHILD")
+
+        # finding non-respite periods of care
+        decom_before_dec_x1 = (df["SW_DECOM"] <= df["DEC"]) & (df["REC"] != "X1")
+        decom_after_dec_s = (df["SW_DECOM"] >= df["DEC"]) & (df["RNE"] != "S")
+        not_V3_V4 = ~df["LS"].isin(["V3", "V4"])
```
SLornieCYC: Think I want the second test here to be `SW_DECOM` >= `DECOM` and `RNE` == "S".  The object name would need renaming then too.

lac_validator/rules/lac2023_24/rule_SW12STG2.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # If any of these are true, the child was not continously looked after
+        not_continous = decom_before_dec_x1 | decom_after_dec_s | not_V3_V4
```
SLornieCYC: Is this actually being used anywhere?  I think rather than "not continuous" this should effectively be the main cohort we care about for this rule - any EPI in this list that doesn't have a SW_DECOM should flag an error for this rule.

lac_validator/rules/lac2023_24/rule_SW12STG2.py
SLornieCYC: Do these two tests need to allow null SW_DECOM values as well?  Otherwise are we only getting the "valid" records?

lac_validator/rules/lac2023_24/rule_SW12STG2.py
```diff
@@ -0,0 +1,79 @@
...
+
+        in_care_at_some_point = df.groupby("CHILD")["DEC_IN_YEAR"].transform("max")
+        not_in_care_at_some_point = df.groupby("CHILD")[
+            "implies_not_continuous"
+        ].transform("max")
+
+        # If CONTINOUSLY LOOKED AFTER == FALSE, there are previous episodes of care
+        epi["CONTINUOUSLY_LOOKED_AFTER"] = (in_care_at_some_point == True) & (
+            not_in_care_at_some_point == False
+        )
```
SLornieCYC: I'm not sure what this section is doing. Nothing links back to the bit above re the non-respite episodes with non-continuous starts and ends.

lac_validator/rules/lac2023_24/rule_SW12STG2.py
```diff
@@ -0,0 +1,79 @@
...
+        df = pd.merge(epi, SWE, how="left", on="CHILD")
+
+        swdecom_before_dec = df[df["SW_DECOM"] < df["DEC"]]["CHILD"].tolist()
+
+        error_rows = df[
+            (df["CONTINUOUSLY_LOOKED_AFTER"] == True)
+            & ~(df["CHILD"].isin(swdecom_before_dec))
+        ]["index"].tolist()
+
+        return {"SWEpisodes": error_rows.tolist()}
```
SLornieCYC: I think at the end here I want the returned list to be non-continuous episode row IDs that don't have a SW_DECOM after the merge and filtering above?

lac_validator/rules/lac2023_24/rule_SW12STG2.py
WillLP-code: I forgot to actually make the slice!

lac_validator/rules/lac2023_24/rule_SW12STG2.py
WillLP-code: So I've made some changes, still very much a WIP, and to get the merge to work I've had to add an episode order ID so SW and standard Episodes to stop having many to many merges, but it makes a lot of assumptions

lac_validator/rules/lac2023_24/rule_SW12STG2.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # This assumes that missing SW Episodes are only ones not added to the end, not middle ones
+        epi.sort_values(["CHILD", "DECOM"], ascending=True, inplace=True)
+        epi["EPI_ID"] = epi.groupby("CHILD").cumcount()
+        SWE.sort_values(["CHILD", "SW_DECOM"], ascending=True, inplace=True)
+        SWE["EPI_ID"] = SWE.groupby("CHILD").cumcount()
+
+        epi = epi[(epi["DEC"] >= collection_start) & (epi["DEC"] <= collection_end)]
+
+        df = pd.merge(epi, SWE, how="left", on=["CHILD", "EPI_ID"])
```
SLornieCYC: This won't work because episodes and SW episodes don't align in that way.  A child could have several changes of worker (multiple SW episodes) during a single episode or have multiple episodes (due to change of placement or legal status) and remain with the same worker throughout.  There's no guarantee that e.g. EPI (3) has any relationship to SWE (3).

lac_validator/rules/lac2023_24/rule_SW12STG2.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # finding non-respite periods of care
+        swdecom_before_dec_x1 = (
+            (df["SW_DECOM"] <= df["DEC"]) | df["SW_DECOM"].isna()
+        ) & (df["REC"] != "X1")
+        swdecom_after_decom_s = (
+            (df["SW_DECOM"] >= df["DECOM"]) | df["SW_DECOM"].isna()
+        ) & (df["RNE"] == "S")
+        not_V3_V4 = ~df["LS"].isin(["V3", "V4"])
```
SLornieCYC: If you want to reduce the volume of data that needs to be processed many-many in the merge, I'd suggest removing the `REC`, `RNE` and `LS` filters from here and applying them directly on the `epi` dataframe before merging it with `SWE`.

lac_validator/rules/lac2023_24/rule_SW12STG2.py
```diff
@@ -0,0 +1,155 @@
...
+        epi["DEC"] = pd.to_datetime(epi["DEC"], format="%d/%m/%Y", errors="coerce")
+        epi["DECOM"] = pd.to_datetime(epi["DECOM"], format="%d/%m/%Y", errors="coerce")
+
+        # This assumes that missing SW Episodes are only ones not added to the end, not middle ones
+        epi.sort_values(["CHILD", "DECOM"], ascending=True, inplace=True)
+        epi["EPI_ID"] = epi.groupby("CHILD").cumcount()
+        SWE.sort_values(["CHILD", "SW_DECOM"], ascending=True, inplace=True)
+        SWE["EPI_ID"] = SWE.groupby("CHILD").cumcount()
+
+        epi = epi[(epi["DEC"] >= collection_start) & (epi["DEC"] <= collection_end)]
```
SLornieCYC: I'd not bother with filtering the `epi` dates at all, tbh.  The return spec is for LAs to only include a single year of episodes anyway, so if someone is including more than that they have bigger issues (which ought to be flagged in another rule somewhere).  Also only testing `DEC` here means we'd miss out any `RNE` == S that doesn't have any end date!

SLornieCYC:
> @SLornieCYC I've made a further attempt at this, trying to take into account what you said. I haven't written a test yet because I wanted to check if this was along the lines of what you had in mind? I'm happy for you to change/completely re-do this if you want to and find you have time

I've added some more feedback and written some test data based on what I **think** should pass and fail!  (But I could be wrong!)

lac_validator/rules/lac2023_24/rule_SW12STG2.py
WillLP-code: I've tried this a number of ways and moving the S, X1, and V3 V4 checks to epi produces really different behaviour in most configurations so I'd say leave it as it is. For instance, index 9 which should fail according to your comments doesn't get through the not_continous slice if epi is filtered according to

`epi[~epi["LS"].isin(["V3", "V4"]) | (epi["RNE"] == "S") | (epi["REC"]) != "X1"]`

SLornieCYC: I've had a go this afternoon and think I've found a combination that works. Attached rather than committed in case it conflicts with any local changes you have on this branch!
[rule_SW12STG2.txt](https://github.com/data-to-insight/csc-validator-be-903/files/15310019/rule_SW12STG2.txt)

WillLP-code:
@SLornieCYC I can't believe how long I've worked at this only for you to manage it in an afternoon with a really elegant solution! Well done! I'm just getting it in and committing it, then, if you approve it, I'll work on getting a new version pushed with the new ingress and the last of these rules done!

SLornieCYC (approved):

PR #752, opened 2023-12-19
24/25 removes rules 217t, 230, 370

### LINKED ISSUE #700: Error codes deleted from 2024/25 [Deleted 2024 v1.0]
The following validation rules are removed in 2024/25 (but still valid in previous years!).  This is due to legal changes that have led to the removal of the H5 and P2 placement codes this year.

- [x] 217t
- [x] 230
- [x] 370

WillLP-code:
closes #700


SLornieCYC (approved):
A nice easy one to review! Assume this will only affect the rule count test only when we roll that forward to 2024 in the summer sometime?

https://github.com/data-to-insight/csc-validator-be-903/blob/5ba851bb8422ac0b6fdad7254fa61aa900a561cf/tests/test_utils.py#L12-L14

PR #754, opened 2023-12-21
SW13STG1 ammended for 24/24

### LINKED ISSUE #704: Error code SW13STG1 [Amended 2024 v1.0]
**Note: This rule was added by DFE in the 2023/24 v1.1 validation check document. At time of writing not implemented (https://github.com/data-to-insight/csc-validator-be-903/issues/634) due being dependent on https://github.com/data-to-insight/csc-validator-be-903/issues/637.**

#### Error code
SW13STG1
#### Description to display
The reason for social worker change is either not valid or has not been entered.
The reason for social worker change field must not be blank, and code must be taken from the list supplied in the DfE children looked after collection guide.
Valid codes are MANAGE, FCONTA, LEFTRL, ORGRST, TSPROC, ABSENC, CHCHAN, PCCHAN, SWDIED, OTHERS.
Codes must all be upper case.
**This code may only be left blank if it the first social worker episode of the reporting period.**
#### DfE suggested coding
`<SW_REASON>` must be provided and be a valid value
For allowed values please see associated code set.

WillLP-code:
closes #704

### DIFF
--- a/lac_validator/rules/lac2024_25/rule_SW13STG1.py
+++ b/lac_validator/rules/lac2024_25/rule_SW13STG1.py
@@ -1,0 +1,125 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="SW13STG1",
+    message="The reason for social worker change is either not valid or has not been entered.",
+    affected_fields=["SW_REASON"],
+)
+def validate(dfs):
+    if "SWEpisodes" not in dfs:
+        return {}
+    else:
+        df = dfs["SWEpisodes"]
+        df["index"] = df.index
+
+        collection_start = dfs["metadata"]["collection_start"]
+
+        valid_reasons = [
+            "MANAGE",
+            "FCONTA",
+            "LEFTRL",
+            "ORGRST",
+            "TSPROC",
+            "ABSENC",
+            "CHCHAN",
+            "PCCHAN",
+            "SWDIED",
+            "OTHERS",
+        ]
+
+        df["SW_DECOM"] = pd.to_datetime(
+            df["SW_DECOM"], format="%d/%m/%Y", errors="coerce"
+        )
+        df["SW_DEC"] = pd.to_datetime(df["SW_DEC"], format="%d/%m/%Y", errors="coerce")
+        collection_start = pd.to_datetime(collection_start, format="%d/%m/%Y")
+
+        df = df.sort_values(["CHILD", "SW_DECOM"], ascending=True)
+
+        last_of_last = df[df["SW_DECOM"] < collection_start].drop_duplicates(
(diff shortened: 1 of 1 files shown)


lac_validator/rules/lac2024_25/rule_SW13STG1.py
```diff
@@ -0,0 +1,125 @@
...
+        first_of_this = df[df["SW_DECOM"] >= collection_start].drop_duplicates(
+            "CHILD", keep="first"
+        )
+
+        joined_df = pd.concat([last_of_last_closed, first_of_this], axis=0)
+
+        joined_df = joined_df.sort_values(
+            ["CHILD", "SW_DECOM"], ascending=True
+        ).drop_duplicates("CHILD", keep="first")
+        allowed_nans = joined_df[joined_df["SW_REASON"].isna()]["index"].tolist()
```
SLornieCYC: I don't understand what any of this is doing?

WillLP-code: _Trying_ to find the first episode of the reporting period, so either one that started last year that didn't close, or the first one opened in the next year if there wasn't an ongoing one. That might not even be needed that was just my first interpretation of the rule based on the amended text description

SLornieCYC: Sorry, finally got around to revisiting this!

I **think** I now understand where you're going with this but also think you may be overcomplicating it.  And not sure that taking the latest `DECOM` per child pre `collection_start` will work once this module contains two years of data (from 2024/25 where this rule is being amended). 

Would something like this (pseudocode) not work?

```
df.sort_values*
index_of_first**
allowed_nans = index_of_firsts[index_of_firsts["SW_REASON"].isna()]["index"].tolist()
```
*I'd add `SW_DEC` here as well as may be overlapping data on `CHILD` and `SW_DECOM` only (the default behaviour of sort_values will put null DEC at the end of each group)
**Essentially first_of_this from your current code but without the `collection_start` filter (to capture all rows in the dataset now we have two years or SW episodes)

WillLP-code: From this I think I really misunderstood the rule! I've made the changes as I think you suggest them, which also means my test was testing wrong, maybe, so I've updated that too!

SLornieCYC: I think your last test there was fine when you were accounting for `collection_year` but now we're using the entire returned file content to accommodate two years of data that test was no longer relevant.

SLornieCYC (approved):
This looks to work the way I interpet the rule description and logic.  Happy to be proved wrong in future once we can compare the outputs of DFE portal and this tool!  (We're still waiting to get access to the SW episodes section for this year's return).

PR #765, opened 2024-04-04
Update rule 370

### LINKED ISSUE #764: Error code 370 [Amended 2023 v1.3]
#### Error code
370
#### Description to display
Child in independent living should be at least ~~15~~ **16**.
Where the placement code is P2 the date at the start of the placement should be the child's ~~15th~~ **16th** birthday or over.  Check that the date of birth and the placement code are correct. If so contact DfE.
#### DfE suggested coding
Where `<PL>` = 'P2' then `<DOB>` should be >= ~~15~~ **16** years prior to `<DECOM>`

SLornieCYC:
Closes #764

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_370.py
+++ b/lac_validator/rules/lac2023_24/rule_370.py
@@ -1,0 +1,67 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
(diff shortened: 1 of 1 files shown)


WillLP-code (approved):
Yep, looks good!

PR #766, opened 2024-04-04
Amend rule 217t

### LINKED ISSUE #763: Error code 217t [Amended 2023 v1.3]
#### Error code
217t
#### Description to display
Placement type 'P2' is not a valid placement type from 28 October 2023
Independent living is not a valid placement from 28 October 2023.
Any independent living placements must have an end date and must have ended on or before **27** ~~28~~ October.
These young people may have been placed in regulated supported accommodation. If so then this accommodation should be recorded as 'H5' and the provider URN should be provided.
#### DfE suggested coding
Where `<PL>` = 'P2'; `<DEC>` must be present and must be <= **27/10/2023** ~~28/10/2023~~

SLornieCYC:
Closes #763


SLornieCYC:
This change is no longer required following a further update from DFE.  See issue for details.

PR #773, opened 2024-04-24
Amend/error503 d

### LINKED ISSUE #772: Error code 503D [Amended 2024 v1.1]
### **Note:** Rule previously implemented from issue #207 but rule has been amended in updated guidance.  Changes highlighted with **bold** and ~~strikethrough~~.  
This is **likely** to be a one-year only change for 2024/25 (due to the introduction of the `<PL>` K3 code which replaces the previous H5 code) with the exception to handle that year-on-year conversion no longer relevant from 2025/26.

#### Error code
503D
#### Description to display
The placement type in the first episode does not match open episode at end of last year.
Placement type in the first episode must be consistent with equivalent details in the last episode of a previous year’s record which was open at the end of the year.
Check details for placement type are correct. If they are correct, are the details in last year’s final episode correct? Were there one or more intervening episodes which have been left out? It may be necessary to enter additional episode details into the system.
#### DfE suggested coding
Where previous collection year’s final episode `<DEC>` not provided then first episode of this year’s collection `<PL>` must be the same as the previous episode
**[Except if previous year `<PL>` = ‘H5’ and current year `<PL>` = ‘K3’]**

WillLP-code:
closes #772 

Also fixes typo in SW16aSTG2 which prevented running rules


lac_validator/rules/rule_utils.py
```diff
@@ -67,12 +67,26 @@ def field_different_from_previous(dfs, field):
...
+                merged_co[this_one].astype(str) != merged_co[pre_one].astype(str)
+            ) & (
+                (merged_co[this_one].astype(str) != "K3")
+                & (merged_co[pre_one].astype(str) != "H5")
+            )
+            err_mask = err_mask & merged_co["DEC_PRE"].isna()
+
+            err_list = merged_co["index"][err_mask].unique().tolist()
+            err_list.sort()
+            return {"Episodes": err_list}
```
SLornieCYC: What you've done here is clever but now I'm wondering if this could be tweaked to handle the variation entirely within this function (`field_different_from_previous`) without needing the optional parameter or changes to the individual rule entry?

For example:

1) Bring in `metadata[collectionYear]`
2) Do the normal `err_mask` filter (handling all field types)
3) Put the special handling after `err_mask` but before setting `err_list` to the unique index
4) Change your new if to `if field =='PLACE' and collectionYear == '2024/25':` with the "true" part additionally filtering `err_mask` to exclude K3 == H5
5) Set `err_list` to the unique index as normal

This I think would be cleaner and mean that if further value equivalency is needed in future years, those can also be handled as separate if statements in the same way?  Rather than potentially resulting in a proliferation of optional parameters between the function and the individual rules.

lac_validator/rules/rule_utils.py
```diff
@@ -70,6 +72,13 @@ def field_different_from_previous(dfs, field):
         err_mask = merged_co[this_one].astype(str) != merged_co[pre_one].astype(str)
         err_mask = err_mask & merged_co["DEC_PRE"].isna()
 
+        if (field == "PLACE") & (collection_year == "2024"):
+            err_mask = (
+                err_mask
+                & (merged_co[this_one].astype(str) != "K3")
+                & (merged_co[pre_one].astype(str) != "H5")
+            )
```
SLornieCYC: Does this need another set of brackets somewhere so that it **only** excludes K3 (in `this_one`) when `pre_one` (for the same row) is H5?  I can't decide how the engine will handle err_mask & expression & expression.

If `this_one` is K3 and `pre_one` is e.g. U6 we still want that to flag as an error, ditto where `pre_one` is H5 and `this_one` is anything other than K3.

lac_validator/rules/lac2024_25/rule_503D.py
SLornieCYC, line 1: We shouldn't need this copy of the rule file at all now if handling the special case entirely within the rule_util function?  EDIT apart from adding the metadata to the rule test...

lac_validator/rules/rule_utils.py
```diff
@@ -41,6 +41,8 @@ def field_different_from_previous(dfs, field):
     if "Episodes" not in dfs or "Episodes_last" not in dfs:
         return {}
     else:
+        collection_year = dfs["metadata"]["collectionYear"]
```
SLornieCYC, line 44: Adding this here does mean metadata/collection year needs to be included in the test data for **all rules** (in all years) that call the `field_different_from_previous` function. i.e. all of the other alphabetised parts of 503.

SLornieCYC (changes requested):

lac_validator/rules/rule_utils.py
SLornieCYC: I think it might need be a negation rather than an addition?  Like this maybe?

```
if (field == "PLACE") & (collection_year == "2024"):
            err_mask = (
                err_mask
                & ~((merged_co[this_one].astype(str) == "K3")
                & (merged_co[pre_one].astype(str) == "H5"))
            )
```

A few additional tests to handle the special case might be useful here.

lac_validator/rules/lac2024_25/rule_503D.py
WillLP-code: For the metadata reason, I suggest we keep it?

SLornieCYC: Hmm in which case I think it would be clearer to:
1) **Only** include 503D in the 2024/25 folder (the one the special case actually applies to).
2) Update the other 503 parts with metadata in the earliest year they exist (you've already done this in 2022/23) but then **omit** them from the later year folders (because there is no change to those rules or their behaviour).
3) Take your comment out of the 2024/25 version of 503D (where the year **does** matter) and instead add it to all of the 503 parts in 2022/23 (where it doesn't):

```
    # For the test, this doesn't affect anything except how the data is handled
    # by the field_different function, so it doesn't matter that the years
    # in the test dfs are not accurate to this collection year
    # As seen in conftest collection year is read in as XXXX/XX and stored in metadata
    # as the first four characters using [:4] in _process_metadata in datastore```

SLornieCYC (approved):
This looks good to me now!

PR #777, opened 2024-05-03
Update/supported accommodation ingress

WillLP-code:
Adds a couple of lines in the ingress in the scpch provider info function to concatenate the new supported accomodation list to the current combined scpch list to be able to check those URNs too, currently the ingress only uploads the first sheet.

It will pass the test correctly to get sent there in the combined scpch check function.

There are tests for the ombined scpch check function and the  scpch provider info function.


lac_validator/ingress.py
```diff
@@ -431,7 +431,36 @@ def scpch_provider_info_table(scpch: UploadedFile):
...
+        scpch_iter = iter(scpch_providers.values())
+        scpch_1, scpch_2 = next(scpch_iter), next(scpch_iter)
+
+        scpch_2.columns = scpch_2.columns.str.lower()
+        scpch_1.columns = scpch_1.columns.str.lower()
+
+        # Supported accomodation has the placement code H5 (I believe)
+        scpch_2["placement code"] = "H5"
+        # This doesn't appear to be in the data currently?
+        scpch_2["deregistration date"] = pd.NA
```
SLornieCYC: Looking at the sample file you've uploaded (I've not seen a "real" file) you can probably use the "Closed date" column from the excel sheet here.

lac_validator/ingress.py
```diff
@@ -431,7 +431,36 @@ def scpch_provider_info_table(scpch: UploadedFile):
...
+        scpch_providers = next(iter(scpch_providers.values()))
+    if len(scpch_providers) == 2:
+        scpch_iter = iter(scpch_providers.values())
+        scpch_1, scpch_2 = next(scpch_iter), next(scpch_iter)
+
+        scpch_2.columns = scpch_2.columns.str.lower()
+        scpch_1.columns = scpch_1.columns.str.lower()
+
+        # Supported accomodation has the placement code H5 (I believe)
+        scpch_2["placement code"] = "H5"
```
SLornieCYC: `H5` for the latter half of 2023/24 only; `K3` for 2024/25 onwards.  Think you can include both here as a comma separated string based on the function notes above re output columns?

lac_validator/ingress.py
```diff
@@ -431,7 +431,36 @@ def scpch_provider_info_table(scpch: UploadedFile):
...
+                "provider local authority",
+                "placement code",
+                "placement provider code",
+                "deregistration date",
+                "setting address postcode",
+            ]
+        ]
+        scpch_providers = pd.concat([scpch_1, scpch_2])
+
     scpch_providers.columns = scpch_providers.columns.str.lower()
```
SLornieCYC: Should this str.lower line be in the `if len(scpch_providers) == 1:` block now?

SLornieCYC (approved):
This looks good to me!

PR #778, opened 2024-05-15
amended Sw12STG for 2024

### LINKED ISSUE #703: Error code SW12STG2 [Amended 2024 v1.0]
**Note: This rule was added by DFE in the 2023/24 v1.1 validation check document. At time of writing not implemented (https://github.com/data-to-insight/csc-validator-be-903/issues/680) due being dependent on https://github.com/data-to-insight/csc-validator-be-903/issues/637.**

#### Error code
SW12STG2
#### Description to display
If a child is looked after on 31 March then all social worker episodes for the full previous **two** year**s** should be reported, even those whilst they were not looked after
The child has had more than one period of care in the year. There is no social worker episode reported for the time prior to the latest period of care.
Social worker episode data should be provided for the full period 1 April to 31 March for any child looked after on 31 March. Note: This includes any periods when the child was not looked after but still had a social worker.
#### DfE suggested coding
For the child, if any previous non-respite `<PERIOD_OF_CARE>` exists where the last `<DEC>` is in the `<CURRENT_COLLECTION_YEAR>` **minus 1 year**
then for each period of care there must be at least one social worker episode where
`<SW_DECOM>` < last `<DEC>` in that `<PERIOD_OF_CARE>`

WillLP-code:
closes #703 

I'm not convinced that this is what the rule logic asks for re: DEC being within the previous collection year but it's not hard to change.

### DIFF
--- a/lac_validator/rules/lac2024_25/rule_SW12STG2.py
+++ b/lac_validator/rules/lac2024_25/rule_SW12STG2.py
@@ -1,0 +1,174 @@
+import pandas as pd
+import numpy as np
+
+from lac_validator.rule_engine import rule_definition
+from lac_validator.utils import add_col_to_episodes_CONTINUOUSLY_LOOKED_AFTER
+
+
+@rule_definition(
+    code="SW12STG2",
+    message="If a child is looked after on 31 March then all social worker episodes for the full previous year should be reported, even those whilst they were not looked after",
+    affected_fields=["SW_DECOM"],
+)
+def validate(dfs):
+    if ("SWEpisodes" not in dfs) | ("Episodes" not in dfs):
+        return {}
+    else:
+        SWE = dfs["SWEpisodes"]
+        epi = dfs["Episodes"]
+
+        collection_end = dfs["metadata"]["collection_end"]
+        collection_end = pd.to_datetime(collection_end, format="%d/%m/%Y")
+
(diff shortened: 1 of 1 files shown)


lac_validator/rules/lac2024_25/rule_SW12STG2.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # The rule states that DEC must be within the previous collection year, but that doesn't include
+        # children who have not ceased an episode yet
+        epi = epi[
+            (
+                (epi["DEC"] >= collection_start - np.timedelta64(1, "Y"))
+                & (epi["DEC"] <= collection_end - np.timedelta64(1, "Y"))
+            )
+            | epi["DEC"].isna()
+        ]
```
SLornieCYC: The error description talks about "full previous two years" (vs "full previous year" in original rule) so I think this needs to capture all `DEC` between `collection_start` -1yr to `collection_end` in the **current year** (and the NAs).

WillLP-code: Can probably even skip collection end then, but I'll keep it in regardless.

SLornieCYC (approved):
I'll approve this so we can get a new version published; we've just got the UAT site for the SWE bit of the current return so hopefully can then do some comparitative testing between D2I tool and the live return.

PR #784, opened 2024-05-22
Add cli command to download info for rules

WillLP-code:
Adds a cli command `rule-info` which allows users to select a year and download all the rules, rule messages, affected fields, and affected tables for those rules as a csv. This was a request from an LA analyst and is supposed to match the same thing which can be downloaded from the real portal. I'll add a download link to the FE later. 

Required adding a tables variable to the decorator for every rule function, which was done by hand.

Future feature: add tables info to the report download after rules are run.


lac_validator/__main__.py
```diff
@@ -36,6 +39,42 @@ def list_cmd(ruleset):
...
+    for _, rule in ruleset_registry.items():
+        rules_list.append(
+            {
+                "Rule code": f"{rule.code}",
+                "Rule message": f"{rule.message}",
+                "Affected fields": f"{sorted(rule.affected_fields)}",
+                "Tables": f"{sorted(rule.tables)}",
+            }
+        )
+    rule_info = pd.DataFrame(rules_list)
```
SLornieCYC, line 69: Tempted to suggest sorting the df by rule code here, before sending it to csv.

lac_validator/__main__.py
```diff
@@ -36,6 +39,42 @@ def list_cmd(ruleset):
...
+    click.echo(rule_info)
+
+    # uncomment to check if any rules are midding table info
+    # click.echo(rule_info[rule_info["Tables"].isna()])
+
+    rule_info.to_csv(f"output_data/903_{ruleset[3:]}_rule_info.csv")
```
SLornieCYC: Is it worth excluding the df index column from this so it doesn't show in the exported csv?  Doesn't really serve any purpose there (and doubly so once someone opens the csv in e.g. Excel or NP+ which independently autonumber rows).

SLornieCYC (commented):
I don't recall ever seeing such a file in the DFE CLA portal but *shrug* I don't suppose that really matters if someone finds this format useful!  Made two minor suggestions below.

lac_validator/__main__.py
WillLP-code: So, I'll do this, but it has to be sorted as strings, not as numbers unless I really want to bloat out this little bit of code, but that does mean that rule 1001 comes before, say, rule 101.

I could maybe do it without too much code but I don't want to spend too much time on something that was just a little side ask from one LA.

lac_validator/__main__.py
WillLP-code: Great shout, I didn't even know this was an option!

SLornieCYC (approved):

PR #786, opened 2024-06-20
fix syntax in publish.yml

tab1tha:
Fixes syntax error that [caused tests to fail ](https://github.com/data-to-insight/csc-validator-be-903/actions/runs/9544870047/workflow)at the merge step.

Related to https://github.com/data-to-insight/csc-validator-be-903/pull/785

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -9,8 +9,8 @@ jobs:
 
     runs-on: ubuntu-latest
     permissions:
-+     # IMPORTANT: this permission is mandatory for trusted publishing
-+     id-token: write
+      # IMPORTANT: this permission is mandatory for trusted publishing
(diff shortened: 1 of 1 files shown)


tab1tha:
Documenting debug process

[This stackoverflow answer](https://stackoverflow.com/a/20420243/11044184) suggested `pyyaml` as a package that would help the developer check, within the local environment, that a yaml file is valid.

To run it on Windows where the `<` symbol might not be recognised, do this: 
- run poetry add pyyaml (or pip install pyyaml to install the package) 
- Create an empty python file with the code below, and run it. Remember to write the file path as a string. 
``` 
import yaml 
print(yaml.safe_load("write_path_to_yaml_file_here")) 
```
If an error is returned, then the file is not properly formatted.

PR #787, opened 2024-06-21
Rules SW11aSTG2 and SW11bSTG2

### LINKED ISSUE #781: Error code SW11aSTG2 [Added 2023 v1.2]
**Note: This rule was added by DFE in the 2023/24 v1.2 validation check document. It replaces rule SW11STG2 (#679)**

#### Error code
SW11aSTG2
#### Description to display
Social worker episode has ended and no subsequent social worker episode has been provided.
The social worker episode has ended.
All social worker episodes should be provided for the time period 1 April previous year to 31 March of the current year for a child looked after on 31 March, including periods when the child was no longer looked after if they still had a social worker,
Missing social worker episodes should be added.
#### DfE suggested coding
If `<SW_DEC>` < (`<COLLECTION_END_DATE>` - 5 days)
then next social worker episode `<SW_DECOM>` must be present and must be = (previous `<SW_DEC>` +/- 5 days)

### LINKED ISSUE #783: Error code SW11bSTG2 [Added 2023 v1.2]
**Note: This rule was added by DFE in the 2023/24 v1.2 validation check document. It replaces SW11STG2 (#679)**

#### Error code
SW11bSTG2
#### Description to display
There should be a social worker episode open on 31 March. Social worker episode has ended before 31 March and no subsequent social worker episode has been provided.
The social worker episode has ended before 31 March and no following social worker episode has been provided.
All social worker episodes should be provided for the time period 1 April to 31 March for a child looked after on 31 March, including periods when the child was no longer looked after if they still had a social worker,
Missing social worker episodes should be added.
If there is no social worker on 31 March then please request a bypass and provide a reason.
#### DfE suggested coding
If `<SW_DEC>` >= (`<COLLECTION_END_DATE>` - 5 days)
then next social worker episode `<SW_DECOM>` must be present

WillLP-code:
closes #781  closes #783 

Had a quick go at these using the original SW11STG2 code, just changing what the error rows look for.

a looks for an SW_DECOM within +/- 5 days of the previous SW_DEC if the SW_DEC is more than 5 days before the end of the collection

b looks for ANY SW_DECOM if the SW_DEC is after 5 days before the end of the collection.

As always, I'm going by the logic, not the words though, so their suggested coding may not match up to what they're really looking for.

### DIFF
--- a/lac_validator/rules/lac2023_24/rule_SW11aSTG2.py
+++ b/lac_validator/rules/lac2023_24/rule_SW11aSTG2.py
@@ -37,15 +37,17 @@ def validate(dfs):
             suffixes=("", "_prev"),
         )
 
-        before_end = m_df[m_df["SW_DEC_prev"] < collection_end - pd.DateOffset(days=5)]
+        before_end = m_df[m_df["SW_DEC_prev"] <= collection_end - pd.DateOffset(days=5)]
         decom_within_timeframe = (
             before_end["SW_DECOM"] >= before_end["SW_DEC_prev"] - pd.DateOffset(days=5)
         ) & (
             before_end["SW_DECOM"] <= before_end["SW_DEC_prev"] + pd.DateOffset(days=5)
         )
 
+        same_child = before_end["CHILD"] == before_end["CHILD_prev"]
+
         error_rows = before_end[
-            before_end["SW_DECOM"].isna() | ~decom_within_timeframe
+            (before_end["SW_DECOM"].isna() | ~decom_within_timeframe) & same_child
         ]["index"]
 
         return {"SWEpisodes": error_rows.tolist()}
--- a/lac_validator/rules/lac2023_24/rule_SW11aSTG2.py
+++ b/lac_validator/rules/lac2023_24/rule_SW11aSTG2.py
@@ -4,7 +4,7 @@ from lac_validator.rule_engine import rule_definition
 
 
 @rule_definition(
-    code="SW11STG2",
+    code="SW11aSTG2",
     message="Social worker episode has ended before 31 March and no subsequent social worker episode has been provided.",
     affected_fields=["SW_DEC"],
     tables=["SWEpisodes"],
@@ -37,11 +37,15 @@ def validate(dfs):
             suffixes=("", "_prev"),
         )
 
-        before_end = m_df[m_df["SW_DEC_prev"] < collection_end]
+        before_end = m_df[m_df["SW_DEC_prev"] < collection_end - pd.DateOffset(days=5)]
+        decom_within_timeframe = (
+            before_end["SW_DECOM"] >= before_end["SW_DEC_prev"] - pd.DateOffset(days=5)
+        ) & (
+            before_end["SW_DECOM"] <= before_end["SW_DEC_prev"] + pd.DateOffset(days=5)
+        )
 
         error_rows = before_end[
-            before_end["SW_DECOM"].isna()
-            | (before_end["SW_DECOM"] != before_end["SW_DEC_prev"])
+            before_end["SW_DECOM"].isna() | ~decom_within_timeframe
         ]["index"]
 
         return {"SWEpisodes": error_rows.tolist()}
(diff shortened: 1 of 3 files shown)


lac_validator/rules/lac2023_24/rule_SW11aSTG2.py
```diff
@@ -0,0 +1,95 @@
...
+        df_lead = df.shift(1)
+        df_lead = df_lead.reset_index()
+
+        m_df = df.merge(
+            df_lead,
+            left_on="index",
+            right_on="level_0",
+            how="outer",
+            suffixes=("", "_prev"),
+        )
```
SLornieCYC: Don't we need to merge on CHILD here as well, to make sure that the episodes we're comparing are actually for the same child?  EDIT: Either than or include CHILD = CHILD_prev in the later expressions.

lac_validator/rules/lac2023_24/rule_SW11bSTG2.py
```diff
@@ -37,12 +37,9 @@ def validate(dfs):
             suffixes=("", "_prev"),
```
SLornieCYC, line 40: Ditto here (l32-38).  Don't we need to merge on CHILD here as well, to make sure that the episodes we're comparing are actually for the same child? EDIT: Either than or include CHILD = CHILD_prev in the later expressions.

lac_validator/rules/lac2023_24/rule_SW11bSTG2.py
```diff
@@ -52,14 +49,27 @@ def test_validate():
...
+                "SW_DECOM": "01/05/2000",
+                "SW_DEC": pd.NA,
+            },  # Passes, rule doesn't suggest a timeframe needed for SW_DECOM
+            {"CHILD": "2", "SW_DECOM": "01/04/2000", "SW_DEC": "25/03/2000"},
             {"CHILD": "2", "SW_DECOM": "02/04/2000", "SW_DEC": "03/04/2000"},
             {"CHILD": "3", "SW_DECOM": "01/04/2000", "SW_DEC": "1/04/2000"},
             {"CHILD": "3", "SW_DECOM": "02/04/2000", "SW_DEC": "03/04/2000"},
             {"CHILD": "4", "SW_DECOM": "01/04/2000", "SW_DEC": "30/03/2000"},
             {"CHILD": "4", "SW_DECOM": "30/03/2000", "SW_DEC": "03/04/2000"},
+            {"CHILD": "5", "SW_DECOM": "01/04/2000", "SW_DEC": "30/03/2000"},
```
SLornieCYC, line 114: I think this line (table index 8) should fail because it ended >= collection_end -5 and there's no subsequent episode?  May be related to other comments about making sure we're comparing for same child.

tests/test_utils.py
```diff
@@ -12,7 +12,7 @@
 def test_get_year_ruleset():
     registry = get_year_ruleset("2024")
 
-    assert len(registry) == 311
+    assert len(registry) == 312
```
SLornieCYC, line 15: Think this still wants to be 311 as we need to remove the originally coded SW11STG2 too.  (These two new rules replaced that one in a later version of the rules from the same return year).  Not sure whether we just delete the SW11STG2 rule file entirely or leave it and add the ID to the deleted rules string for the same year?

SLornieCYC (changes requested):
Looks good but think there are a few issues.

tests/test_utils.py
WillLP-code: So I changed SW11STG2 to SW11aSTG2, and added SW11bSTG2, so I think it makes sense to add one to the rule number? I can add SW11STG2 back and have it in the deleted rules, you're right, but I think that's still 312?

SLornieCYC: Oh, right.  So you did!  All good then!

lac_validator/rules/lac2023_24/rule_SW11aSTG2.py
WillLP-code: good catch, this wasn't in the original rule and I didn't even look at this bit!

lac_validator/rules/lac2023_24/rule_SW11bSTG2.py
WillLP-code: I've been thinking about this, shouldn't LOTS of these rows fail as there are SW_DEC later than collection_end - 5? For istance, child 3 has a SW_DEC on 03/04/2000 (which whilst it's later than the collection period, is after the end - 5 which is what the rule cares about) and no subsequent DECOM?

WillLP-code: If this is true, I have a bunch of changes ready to commit that I think work, and if not, it's back to the grindstone to work out how to find children that fit in with this rule

lac_validator/rules/lac2023_24/rule_SW11bSTG2.py
SLornieCYC: I see you've taken different approaches in the two rules.  On 11a you've used `same_child = before_end["CHILD"] == before_end["CHILD_prev"]` and here in 11b you've included CHILD in the merge.  Obviousy the two methods don't quite return the same thing so we'll have to check (via the test data) which one gets us what we need.

lac_validator/rules/lac2023_24/rule_SW11bSTG2.py
SLornieCYC: I think for child 3 it's actually entirely appropriate for this rule to ignore the rows because both SW_DECOM values are after 31 March (collection_end).  They'd be flagged in another rule as being invalid!

For child 1 (to use another example) I think the first row should technically fail because the subsequent row SW_DECOM is not between the first row SW_DEC (in scope of the collection_end -5) and 31 March, and the second row should be ignored because the SW_DECOM is after 31 March (but again, handled as an error elsewhere).

WillLP-code: Okay so would it be, do you think, appropriate to check for values between collection_end - 5 and collection end?

I'm just trying to puzzle out how we get child 5 to fail, but not others. I'm also not entirely sure the ingress would make a row like the second row for child 6 where they have an NA in both SW_DECOM and SW_DEC, which at the moment is being used to determine if they have an appropriate DECOM

SLornieCYC: Yeah, I think if `SW_DEC_prev` is between `collection_end -5` and `collection_end` then we want the `SW_DECOM` value on the same merged row to be between `SW_DEC_prev` and `collection_end`.

That should work for child 5 based on the outer join merge (actual data on the _prev side and pd.NA on the other?).  For child 6, if a row with two pd.NA dates wouldn't make it through Ingress (and to be fair, even if it did it would be subject to multiple errors!) I think the outer join merge should still return the null line for the comparison as with child 5?

WillLP-code: Okay, with some finagling in the last commit I've found a way to a) include only the rows within collection_end >= SW_DEC >= collection_end - 5 and b) a way to get children who have multiple SW_DEC all within that timeframe, but no following NA SW_DECOM which would be needed. There may still be some edge cases these don't capture though... as there always seem to be!

PR #802, opened 2025-01-14
pin poetry version

tab1tha:
pin poetry version to maintain python version compatibility

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -22,6 +22,8 @@ jobs:
 
     - name: Install Poetry 📜
       uses: snok/install-poetry@v1
+      with:
+        version: 1.8.5
 
     - name: Build package
       run: poetry build


WillLP-code (approved):

SLornieCYC:
Does this version lock need applying in the pyproject.toml file as well?

```[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"```

tab1tha:
Hi @SLornieCYC, 
Nice to meet you again. Happy New Year !

That's probable but since it is not breaking right now, I am inclined to leave it to @WillLP-code to decide. 

For documentation purposes, here are the release notes for Poetry version 2.0 which state that poetry 2.0 and above will no longer be compatible with Python version 3.8 and below 
https://python-poetry.org/blog/announcing-poetry-2.0.0/

image


If the decision is made to update the .toml file, then I think the process will be to 

- [ ] edit the pyproject.toml file to `requires = ["poetry-core=1.8.5"]`
- [ ] run `poetry install` and then `poetry shell `and check that it runs successfully.
- [ ] commit and push the changes.
- [ ] Do the same thing on the cin backend repo.

PR #842, opened 2025-11-20
update sw13stg1 for 24/25

### LINKED ISSUE #814: Error code SW13STG1 [Amended 2024 v1.3]
#### Error code
SW13STG1
#### Description to display
The reason for social worker change is **not valid.** ~~either not valid or has not been entered.~~
The reason for social worker change field must not be blank, and code must be taken from the list supplied in the DfE children looked after collection guide.
Valid codes are MANAGE, FCONTA, LEFTRL, ORGRST, TSPROC, ABSENC, CHCHAN, PCCHAN, SWDIED, OTHERS.
Codes must all be upper case.
This code may only be left blank if it the first social worker episode of the reporting period **and it began before 01/04/2023.**
#### DfE suggested coding
~~`<SW_REASON>` must be provided and be a valid value~~
**If `<SW_REASON>` is present, it must be a valid value.
Else if `<SW_REASON>` is not present then `<SW_DECOM>` must be `< 01/04/2023`**
For allowed values please see associated code set.

WillLP-code:
closes #814 

Allows rows with no SW reason to be valid if their DECOM is before 01/04/2023

### DIFF
--- a/lac_validator/rules/lac2024_25/rule_SW13STG1.py
+++ b/lac_validator/rules/lac2024_25/rule_SW13STG1.py
@@ -5,7 +5,7 @@ from lac_validator.rule_engine import rule_definition
 
 @rule_definition(
     code="SW13STG1",
-    message="The reason for social worker change is either not valid or has not been entered.",
+    message="The reason for social worker change is not valid.",
     affected_fields=["SW_REASON"],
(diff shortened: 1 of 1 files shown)


SLornieCYC (approved):
This looks good to me too.

PR #843, opened 2025-11-27
Update/add dolo rules

WillLP-code:
Adds all the DoLo rules new for the 26/27 return.

This should have been 10 separate PRs, I'm deeply sorry.

### DIFF
--- a/lac_validator/rules/lac2026_27/rule_701.py
+++ b/lac_validator/rules/lac2026_27/rule_701.py
@@ -1,0 +1,46 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="701",
+    message="The end date of the DoLO is not a valid date.",
+    affected_fields=["DOLO_START"],
+    tables=["DoLo"],
+)
+def validate(dfs):
+    if "DoLo" not in dfs:
+        return {}
+
+    dolo = dfs["DoLo"]
+
+    no_empty_ends = dolo[dolo["DOLO_END"].notna()].copy()
+
+    no_empty_ends["DOLO_END_dt"] = pd.to_datetime(
+        no_empty_ends["DOLO_END"], format="%d/%m/%Y", errors="coerce"
+    )
+
+    error_rows = no_empty_ends.index[no_empty_ends["DOLO_END_dt"].isna()]
+
+    return {"DoLo": error_rows.tolist()}
+
+
+def test_validate():
+    import pandas as pd
+
+    fake_data = pd.DataFrame(
+        {
+            "DOLO_END": [
+                "01/01/2020",
+                pd.NA,
+                "2000/12/10",
+            ],
+        }
+    )
+
+    fake_dfs = {"DoLo": fake_data}
+
+    result = validate(fake_dfs)
+
+    assert result == {"DoLo": [2]}
--- a/lac_validator/rules/lac2026_27/rule_701.py
+++ b/lac_validator/rules/lac2026_27/rule_701.py
@@ -6,7 +6,7 @@ from lac_validator.rule_engine import rule_definition
 @rule_definition(
     code="701",
     message="The end date of the DoLO is not a valid date.",
-    affected_fields=["DOLO_START"],
+    affected_fields=["DOLO_END"],
     tables=["DoLo"],
 )
 def validate(dfs):
--- a/lac_validator/rules/lac2026_27/rule_702.py
+++ b/lac_validator/rules/lac2026_27/rule_702.py
@@ -1,0 +1,55 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="702",
+    message="The date that the DoLO ended is before the date that it started.",
+    affected_fields=["DOLO_START"],
+    tables=["DoLo"],
+)
+def validate(dfs):
+    if "DoLo" not in dfs:
+        return {}
+
+    dolo = dfs["DoLo"]
+
+    dolo["DOLO_START_dt"] = pd.to_datetime(
+        dolo["DOLO_START"], format="%d/%m/%Y", errors="coerce"
+    )
+
+    dolo["DOLO_END_dt"] = pd.to_datetime(
+        dolo["DOLO_END"], format="%d/%m/%Y", errors="coerce"
+    )
+
+    error_rows = dolo.index[
+        (dolo["DOLO_END_dt"] <= dolo["DOLO_START_dt"])
+        & (dolo["DOLO_END_dt"].notna())
+        & (dolo["DOLO_START_dt"].notna())
+    ]
+
+    return {"DoLo": error_rows.tolist()}
+
+
+def test_validate():
+    import pandas as pd
+
+    fake_data = pd.DataFrame(
+        {
+            "DOLO_START": [
+                "01/01/2000",
+                "01/01/2000",
+                pd.NA,
+                "02/01/2000",
+                "01/01/2000",
+            ],
+            "DOLO_END": ["01/01/2000", pd.NA, "2000/12/10", "01/01/2000", "02/10/2000"],
+        }
+    )
+
+    fake_dfs = {"DoLo": fake_data}
+
+    result = validate(fake_dfs)
+
+    assert result == {"DoLo": [0, 3]}
--- a/lac_validator/rules/lac2026_27/rule_702.py
+++ b/lac_validator/rules/lac2026_27/rule_702.py
@@ -6,7 +6,7 @@ from lac_validator.rule_engine import rule_definition
 @rule_definition(
     code="702",
     message="The date that the DoLO ended is before the date that it started.",
-    affected_fields=["DOLO_START"],
+    affected_fields=["DOLO_START", "DOLO_END"],
     tables=["DoLo"],
 )
 def validate(dfs):
--- a/lac_validator/rules/lac2026_27/rule_704.py
+++ b/lac_validator/rules/lac2026_27/rule_704.py
@@ -6,7 +6,7 @@ from lac_validator.rule_engine import rule_definition
 @rule_definition(
     code="704",
     message="The date that the DoLO ended is before the start or after the end of the collection year.",
-    affected_fields=["DOLO_START", "DOLO_END"],
+    affected_fields=["DOLO_END"],
     tables=["DoLo"],
 )
 def validate(dfs):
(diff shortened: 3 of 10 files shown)


lac_validator/rules/lac2026_27/rule_701.py
```diff
@@ -0,0 +1,46 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="701",
+    message="The end date of the DoLO is not a valid date.",
+    affected_fields=["DOLO_START"],
```
SLornieCYC: Affected_fields should be DOLO_END here.

lac_validator/rules/lac2026_27/rule_702.py
```diff
@@ -0,0 +1,55 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="702",
+    message="The date that the DoLO ended is before the date that it started.",
+    affected_fields=["DOLO_START"],
```
SLornieCYC: Include DOLO_END in affected_fields too.

lac_validator/rules/lac2026_27/rule_702.py
```diff
@@ -0,0 +1,55 @@
...
+
+    dolo["DOLO_END_dt"] = pd.to_datetime(
+        dolo["DOLO_END"], format="%d/%m/%Y", errors="coerce"
+    )
+
+    error_rows = dolo.index[
+        (dolo["DOLO_END_dt"] <= dolo["DOLO_START_dt"])
+        & (dolo["DOLO_END_dt"].notna())
+        & (dolo["DOLO_START_dt"].notna())
+    ]
```
SLornieCYC, line 30: Collecting error rows could equally be expressed as `DOLO_START >= DOLO_END`.

lac_validator/rules/lac2026_27/rule_704.py
```diff
@@ -0,0 +1,64 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="704",
+    message="The date that the DoLO ended is before the start or after the end of the collection year.",
+    affected_fields=["DOLO_START", "DOLO_END"],
```
SLornieCYC: Don't need DOLO_START here.

lac_validator/rules/lac2026_27/rule_705.py
```diff
@@ -0,0 +1,84 @@
...
+        suffixes=("", "_prev"),
+    )
+
+    same_child = m_dolo["CHILD"] == m_dolo["CHILD_prev"]
+    empty_end_prev = m_dolo["DOLO_END_prev"].isna()
+    dolo_overlap = m_dolo["DOLO_START"] < m_dolo["DOLO_START_prev"]
+
+    error_rows = m_dolo[same_child & (empty_end_prev | dolo_overlap)].sort_values(
+        "index"
+    )
```
SLornieCYC: How are null values in DOLO_END handled for the DF sort?  Doesn't it put them at the end by default?  I'm afraid that cases where the missing end is before a later start won't get captured by this logic? (And not currently covered by test_validate).

E.g.

| CHILD | DOLO_START | DOLO_END |
| --- | --- | --- |
| 1 | 01/04/2025 | 02/04//2025 |
| 1 | 04/04/2025 | |
| 1 | 08/04/2025 | 10/04/2025 |

You can also check [rule 574](https://github.com/data-to-insight/csc-validator-be-903/blob/main/lac_validator/rules/lac2022_23/rule_574.py) for how similar logic has previously been implemented in case that helps identify any additional tests.

SLornieCYC: Should this be LEAD or LAG? If you're sorting by DOLO_END descending then isn't LEAD giving you the **next** end date row rather than the previous one?

lac_validator/rules/lac2026_27/rule_706.py
```diff
@@ -0,0 +1,84 @@
+import pandas as pd
+
+from lac_validator.rule_engine import rule_definition
+
+
+@rule_definition(
+    code="706",
+    message="There is an open DoLO in last year’s return and there is no corresponding DoLO recorded at the start of this year..",
+    affected_fields=["DOLO_START"],
```
SLornieCYC, line 9: Don't think DOLO_START makes sense for affected_fields on this rule.  Use CHILD instead.

lac_validator/rules/lac2026_27/rule_706.py
```diff
@@ -0,0 +1,84 @@
...
+    tables=["DoLo", "DOLO_last"],
+)
+def validate(dfs):
+    if "DoLo" not in dfs:
+        return {}
```
SLornieCYC, line 14: You also need to test whether DoLo_last exists here.  Especially seeing as this rule won't actually _do_ anything until 2027/28 anyway (due to being a new module in 2026/27).

lac_validator/rules/lac2026_27/rule_706.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    # Find rows from this year with no match to last year
+    no_matching_start = m_dolo[m_dolo["index_this"].isna()]
+
+    # We need to return rows from last year with no matching start this year, we can't return rows from this year that
+    # don't match last years as they don't exist!
+    error_rows = no_matching_start["index_last"]
+
+    return {"DoLo": error_rows.to_list()}
```
SLornieCYC, line 38: Will this work?  You're returning the row index from last year's file against the current year's file?  Wont you end up flagging errors against the wrong child/record?

lac_validator/rules/lac2026_27/rule_707.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+@rule_definition(
+    code="707",
+    message="Child ceased to be looked after but there is a DoLO without an end date.",
+    affected_fields=["DOLO_END"],
+    tables=["DoLo", "EPISODES"],
+)
+def validate(dfs):
+    # Uses the logic from 577 which is the same rule but for missing
```
SLornieCYC, line 13: I'm going to assume the logic for this one works without trying to decipher it all again!  Only note is there are a number of comments later on in the script that still refer to the missing table or MIS_ column names etc.

lac_validator/rules/lac2026_27/rule_709.py
```diff
@@ -0,0 +1,130 @@
...
+        return {}
+
+    dolo = dfs["DoLo"]
+    epi = dfs["Episodes"]
+
+    dolo["index_dolo"] = dolo.index
+    epi["index_epi"] = epi.index
+
+    # The rule asks for dolo start and end to be present but if it
+    # didn't we could find overlaps by popping collection end as DOLO_END
```
SLornieCYC, line 27: This is a good point and something I noticed too: as written the rule doesn't capture open DOLOs!  I fully expect DFE to come back with revisions to these rules (or additional rules) later on, but we'll cross that bridge when we get there.

lac_validator/rules/lac2026_27/rule_709.py
```diff
@@ -0,0 +1,130 @@
...
+    epi["DEC_dt"] = pd.to_datetime(epi["DEC"], dayfirst=True, errors="coerce")
+    epi["DECOM_dt"] = pd.to_datetime(epi["DECOM"], dayfirst=True, errors="coerce")
+
+    merged = epi.merge(dolo, on="CHILD", how="inner")
+
+    # We can find overlaps by finding a DOLO_START before a DEC and a DOLO_END after a  after a DECOM
+    overlaps = merged[
+        (merged["DOLO_START_dt"] <= merged["DEC_dt"])
+        & (merged["DOLO_END_dt"] >= merged["DECOM_dt"])
+    ]
```
SLornieCYC, line 46: Similar to above re DOLO_END this omits the latest/current placement (no DEC) from the overlap test.  I'd be inclined to add an `or DEC_dt is null` check against DOLO start here.

lac_validator/rules/lac2026_27/rule_710.py
```diff
@@ -0,0 +1,136 @@
...
+        return {}
+
+    dolo = dfs["DoLo"]
+    epi = dfs["Episodes"]
+
+    dolo["index_dolo"] = dolo.index
+    epi["index_epi"] = epi.index
+
+    # The rule asks for dolo start and end to be present but if it
+    # didn't we could find overlaps by popping collection end as DOLO_END
```
SLornieCYC, line 28: This is a good point and something I noticed too: as written the rule doesn't capture open DOLOs! I fully expect DFE to come back with revisions to these rules (or additional rules) later on, but we'll cross that bridge when we get there.

lac_validator/rules/lac2026_27/rule_710.py
```diff
@@ -0,0 +1,136 @@
...
+    epi["DEC_dt"] = pd.to_datetime(epi["DEC"], dayfirst=True, errors="coerce")
+    epi["DECOM_dt"] = pd.to_datetime(epi["DECOM"], dayfirst=True, errors="coerce")
+
+    merged = epi.merge(dolo, on="CHILD", how="inner")
+
+    # We can find overlaps by finding a DOLO_START before a DEC and a DOLO_END after a  after a DECOM
+    overlaps = merged[
+        (merged["DOLO_START_dt"] <= merged["DEC_dt"])
+        & (merged["DOLO_END_dt"] >= merged["DECOM_dt"])
+    ]
```
SLornieCYC, line 47: Similar to above re DOLO_END this omits the latest/current placement (no DEC) from the overlap test. I'd be inclined to add an `or DEC_dt is null` check against DOLO start here.

SLornieCYC (changes requested):
Lots of rules!  Have a mixture of suggested changes and questions/comments on rule logic below. :)

lac_validator/rules/lac2026_27/rule_709.py
WillLP-code: I've had a go, I'm not sure this is what you meant though!

WillLP-code:
@SLornieCYC I've had a go at all of these. Again, I'm so sorry I put them all in one PR, it was really dumb and very against best practice.

lac_validator/rules/lac2026_27/rule_709.py
SLornieCYC: Yes I think that's exactly what I meant.

SLornieCYC:
> @SLornieCYC I've had a go at all of these. Again, I'm so sorry I put them all in one PR, it was really dumb and very against best practice.

@WillLP-code All of the changes look good here, but I think you missed the feedback on 706.

PR #844, opened 2025-12-02
Fix/383 22 23 and 25 26

### LINKED ISSUE #831: Error code 383 [Amended 2025 v1.2]
#### Error code
383
#### Description to display
A child in a temporary placement must subsequently return to his/her normal placement.
Where the placement code is in the range T0 to T4 the reason for new episode code of the next episode must be P, and the placement code and postcode of the next episode must be the same as the placement code and postcode of the previous one.
A temporary placement, by definition, is one where there is clear expectation from the outset that a child will return to his/her normal placement, and that this expectation is duly fulfilled. 
#### DfE suggested coding
If previous episode `<PL>` = ‘T0’ or ‘T1’ or ‘T2’ or ‘T3’ or ‘T4’ then 
current episode `<RNE>` must = ‘P’ 
AND
`<PL>` must = previous episode -1 `<PL>` 
**AND `<PL_POST>` must = previous episode -1 `<PL_POST>`**

For further info, refer to the DfE validation document [here](https://www.gov.uk/government/publications/children-looked-after-return-2025-to-2026-technical-specifications)

WillLP-code:
closes #831 

Fixes the logic in 383 in the original ruleset and 25/26 so that it returns the correct episode, not the previous episode.

### DIFF
--- a/lac_validator/rules/lac2025_26/rule_383.py
+++ b/lac_validator/rules/lac2025_26/rule_383.py
@@ -37,19 +37,6 @@ def validate(dfs):
         epi, left_on="index", right_on="prev-1_index", suffixes=("", "_prev-1")
     )
 
-    print(
-        epi_2_prev[
-            [
-                "CHILD",
-                "DECOM_dt",
-                "CHILD_prev",
-                "DECOM_dt_prev",
-                "CHILD_prev-1",
-                "DECOM_prev-1",
-            ]
-        ]
-    )
-
     same_child = (epi_2_prev["CHILD"] == epi_2_prev["CHILD_prev"]) & (
         epi_2_prev["CHILD"] == epi_2_prev["CHILD_prev-1"]
     )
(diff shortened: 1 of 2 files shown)


lac_validator/rules/lac2025_26/rule_383.py
```diff
@@ -6,7 +6,7 @@
 @rule_definition(
     code="383",
     message="A child in a temporary placement must subsequently return to his/her normal placement.",
-    affected_fields=["PLACE"],
+    affected_fields=["PLACE", "RNE", "PL_POST"],
```
SLornieCYC, line 9: Can you also add RNE to the 2022_23 version of this rule too?  (But not PL_POST as that's related to the 2025_26 rule change).

SLornieCYC (changes requested):
This looks good!  One minor tweak.

